PR #17, opened 2016-08-22
add support for bookmarks

huangqun:


huangqun:
@mdesiderio @parthshah please check the changes for supporting bookmarks

parthshah:
@mdesiderio - how is removing a bookmark supported? remove it from the links array and patch project with entire bookmarks array (minus the one being removed) ?

mdesiderio:
yes, @parthshah, that is the idea

PR #48, opened 2017-03-04
Lint errors fix

riteshsangwan:


riteshsangwan:
- The modifications to the `.eslintrc` is reverted.
- Fix all the additional errors after reverting the rules configuration.

PR #51, opened 2017-03-21
Feature/search

riteshsangwan:
Projects elasticsearch Implementation.


src/routes/projects/list.js
```diff
@@ -29,64 +29,130 @@ const PROJECT_ATTACHMENT_ATTRIBUTES = _.without(
...
+    fields.projects.push('id');
+  }
+  const searchCriteria = {
+    index: ELASTICSEARCH_INDICES.TC_PROJECT_SERVICE,
+    type: ELASTICSEARCH_INDICES_TYPES.PROJECT,
+    size: criteria.limit,
+    from: criteria.offset,
+    sort: `${order[0]}:${order[1]}`,
+  };
+  let sourceInclude;
```
parthshah: Please move this to a function so we can add unit tests to test different input combinations

src/routes/projects/list.js
```diff
@@ -29,64 +29,130 @@ const PROJECT_ATTACHMENT_ATTRIBUTES = _.without(
...
+    sourceInclude = sourceInclude.concat(_.map(memberFields, single => `members.${single}`));
+  }
+  sourceInclude = sourceInclude.concat(_.map(PROJECT_ATTACHMENT_ATTRIBUTES, single => `attachments.${single}`));
+
+  if (sourceInclude) {
+    searchCriteria._sourceInclude = sourceInclude;        // eslint-disable-line no-underscore-dangle
+  }
+
+  // prepare the elasticsearch filter criteria
+  const boolQuery = [];
```
parthshah: Please wrap searchCriteria functionality into a function so we can add tests for different inputs

src/events/projects/index.js
```diff
@@ -1,76 +1,99 @@
...
-//   });
-// };
+const projectCreatedHandler = (logger, msg, channel) => {
+  const data = JSON.parse(msg.content.toString());
 
+  eClient.create({
+    index: ELASTICSEARCH_INDICES.TC_PROJECT_SERVICE,
+    type: ELASTICSEARCH_INDICES_TYPES.PROJECT,
+    id: data.id,
+    body: data,
```
parthshah: When a project is created the creator is automatically added as a member. We'll need to lookup this member's info (name, email) much like you do in projectMember events before indexing the doc into ES.

src/events/projectMembers/index.js
```diff
@@ -13,97 +21,125 @@ import directProject from '../../services/directProject';
...
+  const httpClient = util.getHttpClient({ id: msg.properties.correlationId });
+
+  const updateESPromise = new Promise((accept, reject) => {
+    const userIds = [urlencode(`userId&${newMember.userId}`, 'utf8')];
+    // fetch the member information
+    httpClient.get(`${config.membersServiceEndpoint}/_search?query=${userIds.join(urlencode(' OR ', 'utf8'))}`)
+      .then((memberDetails) => {
+        let payload = newMember;
+        if (_.has(memberDetails, 'result.content')) {
+          payload = _.merge(newMember, _.pick(memberDetails.result.content,
```
parthshah: `members/_search`  endpoint returns an array of elements. `_.pick` should be called on the first element of this array if one exists. Also please handle error case where a user was not returned. Event should be rejected (nack) in case or such errors after logging appropriately.

riteshsangwan:
The changes requested were complete. See [this](b75f1b862956877fee1c7f3e6afd57a24ee6b125) commit.

PR #56, opened 2017-09-11
Release to allow Method Override Header

vikasrohit:
Supporting overriding HTTP Method via X-HTTP-Method-Override header


vikasrohit:
fyi, @fnisen

PR #61, opened 2018-01-06
Add same regex as client to address validation

chq-matteo:
While we think of a better solution, add a regex to validate the urls
https://github.com/appirio-tech/connect-app/issues/1426
We can use joi to do that
https://github.com/hapijs/joi#example
For the regex see  
https://github.com/appirio-tech/react-components/blob/feature/connectv2/components/Formsy/index.js#L14


chq-matteo:
To verify
0. monitor the browser the requests
1. add a valid link
2. edit the request to some invalid address
3. send the request
4. see if the request is rejected

PR #100, opened 2018-06-26
update auth library

gondzo:
@sachin-maheshwari this can be merged only after creating the v2.4 tag for core auth lib


vikasrohit:
@gondzo can we close this PR? I can see we are already on v2.6 of core library.

gondzo:
yes, this one was duplicate, we have upgraded the library in a separate PR

PR #106, opened 2018-07-10
Feature/project status copilot messages

### LINKED ISSUE #105: Error message for copilots
When fetching projects as a copilot user service returns 403 if copilot is not a member and project is not in reviewed status. We should also return an error message saying `This project is not yet reviewed. Once it is available for pickup you will be able to view and join it.` if status is Draft or InReview, or `Another copilot is already assigned to this project` if status is Active, Completed, Cancelled or Paused

gondzo:
Fixes #105

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,7 @@ workflows:
             - test
           filters:
             branches:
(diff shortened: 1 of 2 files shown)


vikasrohit:
Thanks @gondzo for the quick fix.

PR #110, opened 2018-07-14
Added more bus api events

architectt1:
All events are sent when needed.


src/constants.js
```diff
@@ -61,6 +61,12 @@ export const BUS_API_EVENT = {
   PROJECT_CANCELED: 'notifications.connect.project.canceled',
   PROJECT_ACTIVE: 'notifications.connect.project.active',
 
+  PROJECT_PHASE_TRANSITION_ACTIVE: 'project.phase.transition.active',
```
vikasrohit: @gondzo I think we need to keep `notifications.connect` as prefix for all new events as well.

gondzo: agreed, I'll push the change shortly

vikasrohit:
@gondzo Although, [this](https://github.com/topcoder-platform/tc-project-service/blob/cbfaebf5a280bfc4dd0b721dc6ceb340198b549f/src/events/busApi.js#L369) code was not made in this PR, but it seems this condition is wrong as per inline doc and rationale of throwing the event mentioned afterwards. Could or should we fix it to `_.isEqual(originalWithouDetails, updatedWithouDetails)`?

gondzo:
looks like a bug. agreed on the isEqual change.

vikasrohit:
Could you please do that change in this same PR @gondzo? I haven't checked out the the architectt1's origin yet. Also, seems like their is conflict as well, which needed to be resolved before merging.

After these changes, I think we are safe to merge this PR into dev.

vikasrohit:
@gondzo resolved conflict, merging the PR to reduce the number of things in flight.

PR #112, opened 2018-07-17
Promoting changes to production for connect release 2.4.3

vikasrohit:
It contains following updates to the previous release which provided required features/functionality for connect 2.4.3:

1. project search is improved,
  a. now users can use `ref:<actual_ref_code>` to prioritise the search only on the ref code.
  b. now search respects the project status filter and shows results as per the current selected tab and shows appropriate message to load `All Projects` tab when there is no result for the current selected status tab.
  c. now search respects the `My Projects` switch as well, earlier it was not showing the search results when `My Projects` switch was turned on.
2. Fixed issues with endpoint for migrating old v2 projects to v3 format
3. Fixed admin endpoint for reindexing a project to fetch and index phases
4. Opened GET and LIST endpoints for template entities so that logged out users can create project in connect.
5. Strengthened the write endpoints for the template entities to only allow admins to perform the operations. Earlier managers were also allowed to do these operations.
6. Improved error message for the case where copilot is trying to access already claimed or non ready project.


vikasrohit:
fyi, @mtwomey @rmjmccormick @RishiRajSahu @vic-topcoder 

Deploying project service ahead of our connect release so that can devote enough time for the connect related changes tomorrow. We need sufficient time tomorrow for the release to update the actual templates before going live.

Issue #132, opened 2018-07-27
[$40] Project event links in bus event


gondzo:
All project related bus events should contain the connect project url. Url format is https://connect.topcoder.com/projects/{projectId}. Domain should be read from config (topcoder.com or topcoder-dev.com)

change should be made at https://github.com/topcoder-platform/tc-project-service/blob/dev/src/events/busApi.js

architectt1:
https://github.com/topcoder-platform/tc-project-service/pull/136

vikasrohit:
@coderReview merged the PR.

PR #136, opened 2018-07-28
Issue #132

architectt1:
- Added projectUrl to all connect bus events.
- Added connectProjectsUrl config var to test.json, as it's needed for running tests.

### DIFF
--- a/config/test.json
+++ b/config/test.json
@@ -12,6 +12,7 @@
       "timelineDocType": "timelineV4"
     },
     "rabbitmqUrl": "amqp://localhost:5672",
+    "connectProjectsUrl": "https://local.topcoder-dev.com/projects/",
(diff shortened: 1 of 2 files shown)


config/test.json
```diff
@@ -12,6 +12,7 @@
       "timelineDocType": "timelineV4"
     },
     "rabbitmqUrl": "amqp://localhost:5672",
+    "connectProjectsUrl": "https://local.topcoder-dev.com/projects/",
```
vikasrohit: @architectt1 can you please add `connectProjectsUrl` to other configs as well, so that we don't have to specify it in environment variable.

vikasrohit: I think your other PR contains those changes. Accepting the PR.

Issue #131, opened 2018-07-27
[$40] Add new file url to bus api event


gondzo:
Bus event for uploaded files should contain the file url. Url format is `https://connect.topcoder.com/projects/{projectId}/attachments/{attachmentId}`. Domain should be read from config (topcoder.com or topcoder-dev.com)

chenge should be made at https://github.com/topcoder-platform/tc-project-service/blob/1602c6c53686642b8d08dc440b1a825bb4b60e63/src/events/busApi.js#L170

architectt1:
https://github.com/topcoder-platform/tc-project-service/pull/137

PR #137, opened 2018-07-28
Issue #131

architectt1:
- Added fileUrl when sending a PROJECT_FILE_UPLOADED event to the bus api.
- Added connectProjectsUrl config var to test.json, as it's needed for running tests.

### DIFF
--- a/config/test.json
+++ b/config/test.json
@@ -12,6 +12,7 @@
       "timelineDocType": "timelineV4"
     },
     "rabbitmqUrl": "amqp://localhost:5672",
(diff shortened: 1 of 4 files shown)


config/test.json
```diff
@@ -12,6 +12,7 @@
       "timelineDocType": "timelineV4"
     },
     "rabbitmqUrl": "amqp://localhost:5672",
+    "connectProjectsUrl": "https://local.topcoder-dev.com/projects/",
```
vikasrohit: Can we set default values in all other configs as well? That would make setting this as env variable optional.

architectt1: Sure, done in 84b9fed.

Issue #130, opened 2018-07-27
[$40] Setting StartDate for default and explicitly added phases


RishiRajSahu:
Currently, we don't set start date of a phase during creation. We need to persist start date as well during creation of phase whether its default phase for a project or its a newly added phase.

fyi @vikasrohit @rmjmccormick @acshields 

**Requirements**:
1. Update `POST /v4/projects` endpoint to set the current date as phase `startDate` and set the `endDate` as `startDate + duration`.
2. Update unit tests to validate the startDate and endDate after project creation

vikasrohit:
@RishiRajSahu we won't need to update the `add phase` endpoint because at that its the caller who sends the data for the phase, so caller can send the dates as required.

architectt1:
https://github.com/topcoder-platform/tc-project-service/pull/138

vikasrohit:
Working fine for the specified requirements.

PR #138, opened 2018-07-28
Issue #130

architectt1:
When creating a project from a project template, define phases' `startDate`, `endDate` and `duration`.

### DIFF
--- a/src/routes/projects/create.js
+++ b/src/routes/projects/create.js
@@ -4,6 +4,7 @@ import validate from 'express-validation';
 import _ from 'lodash';
 import Joi from 'joi';
 import config from 'config';
+import moment from 'moment';
 
 import models from '../../models';
 import { PROJECT_MEMBER_ROLE, PROJECT_STATUS, PROJECT_PHASE_STATUS, USER_ROLE, EVENT, REGEX } from '../../constants';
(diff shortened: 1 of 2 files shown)


src/routes/projects/create.js
```diff
@@ -90,12 +91,16 @@ function createProjectAndPhases(req, project, projectTemplate, productTemplates)
...
+      const duration = _.get(phase, 'duration', 1);
+      const startDate = moment.utc();
       // Create phase
-      models.ProjectPhase.create({
+      return models.ProjectPhase.create({
         projectId: newProject.id,
         name: _.get(phase, 'name', `Stage ${phaseIdx}`),
-        duration: _.get(phase, 'duration', 0),
+        duration,
+        startDate: startDate.format(),
```
vikasrohit: @architectt1 Can we make sure that any date field does not persist time of the day in database. We kept the field to be datetime for future, but don't want to add the time details for date fields. I mean, it should always represent 00:00:00.000 in database.

vikasrohit: Assertion in unit test would be great to validate this.

architectt1: @vikasrohit done in de0d52e.

vikasrohit: Thanks @architectt1

Issue #127, opened 2018-07-25
[$60] Updating a milestone's date/duration should update the dates of subsequent milestones


vikasrohit:
As of now, when startDate/endDate/completionDate or duration is updated, we are not cascading the effect to upcoming milestones. Purpose of this ticket is to update the `startDate` and `endDate` of subsequent milestones when we receive call to update `startDate`/`endDate`/`completionDate`/`duration` of any milestone.

fyi @RishiRajSahu @coderReview 

**UPDATE**
It should throw error (422), when the milestone's `startDate` or `endDate` is updated. We won't allow updating `startDate` and `endDate` of milestones.

vikasrohit:
@architectt1 Yes, your assumption is correct in the first para. Thanks for catching that. I missed to consider the scenario when `completionDate` is set for a milestone.

Your point is valid for order change of milestone as well. I am not sure if we are going to use the order change feature ever but we kept it for the sake of comprehensiveness. If we can update the dates of the milestones when their order is changed and that does not leave any use case, we should be good.

For the price point, we can keep the last requirement (about updating dates when order is changed) in separate issue to keep the scope of the issue as per original plan.

architectt1:
@vikasrohit Could we add that to the scope of this issue and increase the price? I ask this because it's hard for me to come up with a *half solution* that doesn't take the order into account. The scenario where I'm wondering what to do is the following:

```
Milestone 1, order 1: 1-3, duration 3, completion date: 4.
Milestone 2, order 2: 5-7, duration 3, completion date: 7.
Milestone 3, order 3: 8-11, duration 4, no completion date.
```

The input, although maybe wrong, we still need to handle it:

PATCH milestone 3, set order to 2, and completion date to 10.

```
Milestone 1, order 1: 1-3, duration 3, completion date: 4.
Milestone 3, order 2 (previously 3): 8-11, duration 4, completion date: 10.
Milestone 2, order 3 (previously 2): 11-13, duration 3, completion date: (previously 7) ???
```

Does that make sense? We can update the start date and end date of the milestone being updated and following ones by code, but if a completion date was already set for later milestones, it doesn't feel ok to change that, nor it doesn't leaving it in an inconsistent state (i.e., leaving it at `7` when the previous milestone was updated to `10`).

When writing this I see that it could also apply to a scenario where order didn't change, but I still think that as I said before, *it's hard for me to come up with a *half solution* that doesn't take the order into account*.

coderReview:
[...]

PR #144, opened 2018-07-28
Issue #127

architectt1:
- Updating a milestone's completionDate/duration should update the dates of subsequent milestones.
- Fixed bug where the transaction wasn't being included into wrapped queries.

### DIFF
--- a/src/routes/milestones/update.js
+++ b/src/routes/milestones/update.js
@@ -3,6 +3,7 @@
  */
 import validate from 'express-validation';
 import _ from 'lodash';
+import moment from 'moment';
 import Joi from 'joi';
 import Sequelize from 'sequelize';
 import { middleware as tcMiddleware } from 'tc-core-library-js';
@@ -12,6 +13,37 @@ import { EVENT } from '../../constants';
 import models from '../../models';
 
 const permissions = tcMiddleware.permissions;
+
+/**
+ * Cascades endDate/completionDate changes to all milestones with a greater order than the given one.
+ * @param {Object} updatedMilestone the milestone that was updated
+ * @param {Object} transaction the wrapping transaction
+ * @returns {Promise<void>} a promise
+ */
+async function updateComingMilestones(updatedMilestone, transaction) {
+  const comingMilestones = _.sortBy(await models.Milestone.findAll({
+    where: {
+      timelineId: updatedMilestone.timelineId,
+      order: { $gt: updatedMilestone.order },
+    },
+    transaction,
+  }), 'order');
+  let startDate = moment.utc(updatedMilestone.completionDate
+    ? updatedMilestone.completionDate
+    : updatedMilestone.endDate).add(1, 'days').toDate();
+  const promises = _.map(comingMilestones, (_milestone) => {
+    const milestone = _milestone;
+    if (milestone.startDate.getTime() !== startDate.getTime()) {
+      milestone.startDate = startDate;
+      milestone.endDate = moment.utc(startDate).add(milestone.duration - 1, 'days').toDate();
+    }
+    startDate = moment.utc(milestone.completionDate
+      ? milestone.completionDate
+      : milestone.endDate).add(1, 'days').toDate();
+    return milestone.save({ transaction });
+  });
+  await Promise.all(promises);
+}
 
 const schema = {
   params: {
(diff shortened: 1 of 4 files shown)


vikasrohit:
@architectt1 I think we don't need to include transaction explicitly. It is automatically handled by [continuation-local-storage](http://docs.sequelizejs.com/manual/tutorial/transactions.html#automatically-pass-transactions-to-all-queries).
Please remove explicit passing of transaction. I think winner of the challenge, where we got timeline/milestone developed, also added some unit tests to verify that transaction is actually working. Let me know if you are seeing otherwise.
fyi @coderReview

src/routes/milestones/update.spec.js
```diff
@@ -898,6 +876,76 @@ describe('UPDATE Milestone', () => {
...
+        .send({ param: _.assign({}, body.param, {
+          completionDate: '2018-05-18T00:00:00.000Z', order: undefined, duration: undefined,
+        }) })
+        .expect(200)
+        .end(() => {
+          // Milestone 3: startDate: '2018-05-14T00:00:00.000Z' to '2018-05-19T00:00:00.000Z'
+          //                endDate: null                       to '2018-05-21T00:00:00.000Z'
+          // Milestone 4: startDate: '2018-05-14T00:00:00.000Z' to '2018-05-22T00:00:00.000Z'
+          //                endDate: null                       to '2018-05-24T00:00:00.000Z'
+          setTimeout(() => {
```
vikasrohit: I don't think we need timeout call here as it was needed only when there was a bug in the way we were handling the transactions i.e. HTTP response was being sent before transaction commit. I fixed those issues with `feature/timeline-milestone` branch.

src/routes/milestones/update.spec.js
```diff
@@ -898,6 +876,76 @@ describe('UPDATE Milestone', () => {
...
+          Authorization: `Bearer ${testUtil.jwts.admin}`,
+        })
+        .send({ param: _.assign({}, body.param, { duration: 5, order: undefined, completionDate: undefined }) })
+        .expect(200)
+        .end(() => {
+          // Milestone 3: startDate: '2018-05-14T00:00:00.000Z' to '2018-05-19T00:00:00.000Z'
+          //                endDate: null                       to '2018-05-21T00:00:00.000Z'
+          // Milestone 4: startDate: '2018-05-14T00:00:00.000Z' to '2018-05-22T00:00:00.000Z'
+          //                endDate: null                       to '2018-05-24T00:00:00.000Z'
+          setTimeout(() => {
```
vikasrohit: Same here about need of timeout

src/routes/milestones/update.spec.js
```diff
@@ -482,28 +480,10 @@ describe('UPDATE Milestone', () => {
         .expect(422, done);
     });
 
-    it('should return 422 if startDate is after completionDate', (done) => {
+    it('should return 422 if startDate is different than the original startDate', (done) => {
```
vikasrohit: Assuming changes for https://github.com/topcoder-platform/tc-project-service/issues/148 in the same PR, I think now it makes more sense to just not accept the `startDate` and `endDate` in payload instead of validating only for different value in these fields.

src/routes/milestones/update.spec.js
architectt1: Done in 737c0e6.

src/routes/milestones/update.spec.js
architectt1: Done in 737c0e6.

src/routes/milestones/update.spec.js
architectt1: Done in 737c0e6.

src/routes/milestones/update.js
```diff
@@ -146,6 +179,16 @@ module.exports = [
                 },
               });
             });
+        })
+        .then(() => {
+          // Update dates of the other milestones only if the completionDate nor the duration changed
+          if (((!original.completionDate && !updated.completionDate) ||
```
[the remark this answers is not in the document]
architectt1: Yes this was an attempt of applying the `avoid else, return early` principle, but since there's a single line next, I'm going to apply the change you recommend.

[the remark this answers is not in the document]
architectt1: Done in 61cdff3.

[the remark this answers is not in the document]
architectt1: And made it more readable as well :)

architectt1:
https://github.com/topcoder-platform/tc-project-service/issues/148 changes are present in 377b070.

architectt1:
@vikasrohit no problem.

Regarding your question, I forgot to remove that from the description, that was from before when I thought that was needed. I just removed that.

Issue #129, opened 2018-07-26
[$100] Need new endpoint to copy milestone templates from one template to another


vikasrohit:
**New endpoint** : POST /v4/productTemplates/{templateId}/milestones/clone
**Payload**:
```
{ param: { sourceTemplateId: 123 } }
```
It should clone all the template from template identified by `sourceTemplateId` to the template identified by `templateId`.

**Deliverables**: PR with following changes
1. Updated source code
2. Unit tests for the new endpoint
3. Updated Swagger doc
4. Updated Postman collection

fyi @coderReview @RishiRajSahu

mhikeo:
Hi. I cannot assign it to myself yet but I'm taking this up. Thanks.

coderReview:
@mhikeo how is it going? when do you plan to submit?

mhikeo:
@coderReview Submitting within the hour.

vikasrohit:
@coderReview @mhikeo merging the PR for now to reduce the possibilities of conflicts with other fixes. I have made request of small change in the PR which is needed to be worked on before we close this issue. We will verify the changes in dev env and there after will close the issue if found working.

PR #146, opened 2018-07-29
Fix for Issue #129

mhikeo:


src/routes/milestoneTemplates/clone.js
```diff
@@ -0,0 +1,100 @@
...
+  },
+};
+
+module.exports = [
+  validate(schema),
+  permissions('milestoneTemplate.clone'),
+  (req, res, next) => {
+    let result;
+
+    return models.sequelize.transaction(tx =>
```
vikasrohit: we can use implicit transactions as other endpoints in the timeline and milestone endpoints are using. This would make code consistent and would not confuse developers who work on them in future.

Issue #140, opened 2018-07-28
[$30] GET /v4/projects endpoint always returns the attachments data


vikasrohit:
[list.js](https://github.com/topcoder-platform/tc-project-service/blob/dev/src/routes/projects/list.js#L137) should not return attachments data if not requested as [list-db.js](https://github.com/topcoder-platform/tc-project-service/blob/dev/src/routes/projects/list-db.js#L61) is doing.

Should update the unit tests accordingly.

vikasrohit:
Please accept the invitation, once done I will assign you.

narekcat:
@vikasrohit I accepted the invitation, thank you.

narekcat:
@vikasrohit I setup application, and try to create project, but there are problem with creating direct project (https://api.topcoder-dev.com/v3/direct/projects/). It responses with status code 401 and message "Valid credentials are required to access this resource.". I try with Bearer code provided in README.md and also try with Bearer code in default.json, bet result is same. Are there any problems with direct project service endpoint ? Are this Bearers wrong ?

vikasrohit:
You can skip direct call. Either resolve empty promise instead of calling
direct or remove that block for your local setup.

narekcat:
@coderReview I use `feature/timeline-milestone` branch, but ok I'll do like @vikasrohit said. And one more thing. Today I have setted up application, this was first time that I were working on this project, that's way setup takes more time. The last issue that I were working on is issue with direct project endpoint. It took about 3 hours to find it out and now I need more time to finish bug. Can I work on it a little longer ?

PR #149, opened 2018-07-29
Issue #140 GET /v4/projects endpoint always returns the attachments data

narekcat:
#140 GET /v4/projects endpoint always returns the attachments data

### DIFF
--- a/src/routes/projects/list.js
+++ b/src/routes/projects/list.js
@@ -134,7 +134,10 @@ const parseElasticSearchCriteria = (criteria, fields, order) => {
     const phaseFields = _.get(fields, 'project_phases_products');
     sourceInclude = sourceInclude.concat(_.map(phaseFields, single => `phases.products.${single}`));
   }
-  sourceInclude = sourceInclude.concat(_.map(PROJECT_ATTACHMENT_ATTRIBUTES, single => `attachments.${single}`));
+  if (_.get(fields, 'attachments', null)) {
+    const phaseFields = _.get(fields, 'attachments');
+    sourceInclude = sourceInclude.concat(_.map(phaseFields, single => `attachments.${single}`));
+  }
 
   if (sourceInclude) {
     searchCriteria._sourceInclude = sourceInclude;        // eslint-disable-line no-underscore-dangle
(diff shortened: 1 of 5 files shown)


src/routes/projects/list.js
```diff
@@ -134,7 +134,10 @@ const parseElasticSearchCriteria = (criteria, fields, order) => {
     const phaseFields = _.get(fields, 'project_phases_products');
     sourceInclude = sourceInclude.concat(_.map(phaseFields, single => `phases.products.${single}`));
   }
-  sourceInclude = sourceInclude.concat(_.map(PROJECT_ATTACHMENT_ATTRIBUTES, single => `attachments.${single}`));
+  if (_.get(fields, 'attachments', null)) {
+    const phaseFields = _.get(fields, 'attachments');
```
vikasrohit: I think copy paste magic is here. We need to rename the `phaseFields`.

src/routes/projects/list.spec.js
```diff
@@ -306,6 +306,83 @@ describe('LIST Project', () => {
...
+          .expect('Content-Type', /json/)
+          .expect(200)
+          .end((err, res) => {
+            if (err) {
+              done(err);
+            } else {
+              const resJson = res.body.result.content;
+              should.exist(resJson);
+              resJson.should.have.lengthOf(3);
+              resJson[0].should.have.property('attachments');
```
vikasrohit: It would be great if you can validate the fields of attachment objects as well, in one unit test.

vikasrohit:
@narekcat I am not sure if our challenge guideline mentioned this or not, but for future bug bash challenges, please create a fork of the repo and create PR from your fork to the origin.

src/routes/projects/list.js
narekcat: Yes, it is really copy paste magic.

narekcat:
@vikasrohit In the challenge guideline didn't mention, but for future challenges I'll create fork. And I'll try to squash changes in one commit for one issue.

src/routes/projects/list.spec.js
narekcat: Ok, I'll do.

narekcat: @vikasrohit I have fixed this comment too.

vikasrohit:
Thanks @narekcat. I would like to have the assertions for values of the fields instead of just checking existence of the properties. However, I think might go out of scope of original requirements, hence, accepting the PR. Will create a separate issue for adding field value validations.

@coderReview merging PR to unblock other PRs and resolve conflicts with other PRs. We can close the issue once we verify the changes in dev.

narekcat:
@vikasrohit Sorry for that, but you wrote to add validations for attachments fields only. Ok, it isn't a problem, I'll add assertions for values of fields when you open issue.

coderReview:
I will update wiki

Issue #126, opened 2018-07-25
[$40] Update PATCH timeline endpoint to update dates of milestones


vikasrohit:
When user is updating the start date of a timeline using the `PATCH /v4/timeline/123` endpoint, it should also update dates(startDate and endDate) of all milestones for that timeline. It should start the first milestone (identified via `order=1`) with timelines start date and calculate end date of that milestone using duration and update the subsequent milestones in the same way.

fyi @coderReview

architectt1:
@coderReview @vikasrohit Are milestones of a given timeline supposed to be one after the other? for example:

* Milestone 1: day 1-3.
* Milestone 2: day 4-7.
* Milestone 3: day 8-10.

And given this example, would change the `startDate` of the timeline to day 5 make the above:

* Milestone 1: day 5-7.
* Milestone 2: day 8-11.
* Milestone 3: day 12-14.

?

(I know these are actually dates, but I'm putting an easy-to-follow example with numbers).

architectt1:
Answering to myself, I believe this is correct, as there's some code doing exactly that when creating milestones from `ProductMilestoneTemplate` records.

vikasrohit:
Correct. Our backend should always ensure that milestones dates are in sync as per their `order`.

architectt1:
Got it. One thing I'm going to rely on is in that the data is correct in the database. For example there might be milestones that step into each other. I.e.

* Milestone 1: day 5-7.
* Milestone 2: day 6-8.
* Milestone 3: day 7-10.

For all cases I'm just going to add the timeline difference to the start date, and then set the end date to be milestone start date + milestone duration. I believe this is a fair approach to use.

architectt1:
Hmm actually because of a test I saw, even if a milestone has a startDate and a duration, it can have an end date not set. Apparently this should stay as that, so in those scenarios I'll just update the startDate.

architectt1:
Do you mean the dates stepping into each other, or the null endDate, or both? the endDate column is marked as nullable in the database. Maybe I should always set the endDate, even when null (= startDate + duration)?

vikasrohit:
Right now I can not collect any use case of having null `endDate`, Yes please, set `endDate` even if it is null.

architectt1:
https://github.com/topcoder-platform/tc-project-service/pull/139

vikasrohit:
@coderReview It is working fine. Please close the issue and mark it for payment.

coderReview:
[...]

PR #153, opened 2018-07-30
Issue #126

gfzabarino:
- Fixed issue where identical data were being treated as different and then triggering milestones' updates.
- Do not trigger milestones' updates if the timeline endDate changed.
- Added comments to milestones' updates when a timeline gets updated.


architectt1:
Please disregard.

Issue #172, opened 2018-08-24
New bus events for milestone notifications


vikasrohit:
Need to support few new events for milestones similar to what we did for project plan earlier in https://github.com/topcoder-platform/tc-project-service/pull/110.

Current list of events is:

`notifications.connect.project.phase.milestone.transition.active` => when milestone is marked as active
`notifications.connect.project.phase.milestone.transition.completed` => when milestone is marked as complete
`notifications.connect.project.phase.milestone.waiting.customer` => when milestones are waiting for customers's input (this might be tricky to identify because every milestone type has different way to identify that it is waiting for customer input and that is stored in `details` json)

ngoctay:
@vikasrohit @gondzo 
I will make a commit for the first 2 events.
For the last one (waiting.customer), will leave a TODO in the code so that we can tackle it later when we have more details.

vikasrohit:
@ngoctay can you please raise the `...waiting.customer` event when we find `details.metadata.waitingForCustomer` flag to exist and equal to `true`? As of now, we don't have a unique way to identify that a milestone is waiting for customer input, so we are thinking of keeping a new flag at mentioned level to capture this details. This field would be updated by front end. Ideally, we don't want to parse the `details` field in backend but I think we are good in this case as it would be a standard flag across milestones and we would keep `details.content` as milestone specific data which we would never parse on backend.

vikasrohit:
It was not working for `active` event when the milestone is made active as cascaded effect after marking previous milestone as `completed` and this was happening because we were using `cascadedUpdates` instead of `cascadedUpdates.milestones` to iterate over milestones updated via cascade. After fixing that it is working fine now for `active` and `completed` events,  however, we need to do some changes in front end to validate the `waiting.customer` event.
Marking the issue closed. Will take care of `waiting.customer` event separately after front end changes.
@gondzo please feel free to tag it as per convenience for payments.

Issue #171, opened 2018-08-24
New authorization scheme for project creation endpoint


vikasrohit:
To support a new use case, where we need to create projects for inactive users, we need to implement a new auth scheme in project service and use that only for project creation endpoint to allow the creation of projects by inactive users as well.

Details:
1. We will look for `Authorization` header
2. Technically it would still be using `Bearer` auth scheme but its value would not be a jwt token this time, rather it would be a simple string with following pattern like `userId_<userId>` i.e. `useId_` as prefix to the actual `userId`.
3. if we find the bearer token starting with `userId_`, we should not use the jwt auth middle ware from `tc-core-library-js` to validate the token and rather use custom logic.
4. if the passed `userId` is inactive (we can query that using admin access to the identity api), we should authorize the user and if `userId` is active, we need to throw 403 with appropriate message.

ngoctay:
@vikasrohit 
1. Only for project creation endpoint, right?
2. "we can query that using admin access to the identity api", can you provide more details on on how to access the identity API?

gondzo:
@vikasrohit does identity api support m2m tokens? do we have a clientId/secret that we could use?

@ngoctay I think identity service endpoint is v4_api/users. Is this correct @vikasrohit ?

vikasrohit:
@ngoctay [this](https://github.com/topcoder-platform/tc-project-service/blob/dev/src/util.js#L294) should be the helper function you should be using to get user's status.

vikasrohit:
@gondzo I think m2m are now supported by identity, however, if our current api is not using m2m to get system user token, we can still go with the existing method of clientId and secret.

ngoctay:
@vikasrohit 
Thanks for the helper function.

Is there any dev identity API endpoint so that I can set the env?

vikasrohit:
Yes, there is. It should be `https://api.topcoder-dev.com/v3`, however, it won't work on your local because you won't have client id and secret. You can test this on local by mocking the endpoint while unit testing.

ngoctay:
Do you have any Swagger or Postman for the Identity API?
I don't know any about the request/response.

vikasrohit:
@ngoctay this should help.
```
getStub.withArgs(`${config.get('identityServiceEndpoint')}users`, getUsersOpts).resolves({
  data: { result: { status: 200, content: [{ id: 1, ..., active: false }] } },
});
```

ngoctay:
@vikasrohit 
[...]

Issue #169, opened 2018-08-22
Add new field `metadata` to the `milestoneTemplate` model


vikasrohit:
This new field would be of type `json` which would serve as storing milestone's type specific configuration e.g. for `final-designs` milestone type, we have to store configuration for the number of final designs that have to be selected to move further.

Scope of this tasks is to:
1. add new field `metadata` to the model
2. accept this new model in create endpoint. it would save as field as is.
3. accept this new model in update endpoint. It would merge the incoming json with existing json as follows (Please see the implementation in `projectTemplate` update endpoint for reference):
  3.1 It should merge non array fields
  3.2 override the array fields with incoming value
  3.3 delete the null valued keys
4. Update/Add unit tests to test the new field and its update logic
5. Update postman and swagger docs

vikasrohit:
Closing as working fine as per given requirements.

@gondzo Please feel free to tag it as per convenience for payment.

Issue #168, opened 2018-08-21
Add support for `order` in phases


vikasrohit:
We need to support the `order` for each phase. Initially, it would be an optional field and would be set only for new phases created after this change.
Cases where we need to update the `order` of the other phases in the same project:

1. when a phase has change in `order` field while updating a phase 
2. when a new phase is added with `order` field and there exists at least one phase at or after that `order`

Also, do let us know if you think there is another case(s) where we need to cascade the changes of the phase's order. We are doing similar stuff for `milestones` and `milestoneTemplates`

PR #174, opened 2018-08-28
Release to support Connect 2.4.5

ngoctay:
#172 #171 #169 #168 #163 #161 #151 #157


ngoctay:
@vikasrohit @gondzo 
Please review.
This includes all issues except #162

vikasrohit:
Thanks @ngoctay. Is it possible for you to create separate PRs for different issues? Actually, single PR makes things complicated when merging them because it would introduce all changes at once and also makes reverting of one particular fix if there is a problem with that.
If it is not more than 15-30 minutes for you, then please separate it out, otherwise, it is okay for this time, but in future please try to create separate PRs from different feature branches in your fork. Thanks for the hard work!!!

src/events/busApi.js
```diff
@@ -414,4 +415,92 @@ module.exports = (app, logger) => {
...
+      }
+
+      if (event) {
+        createEvent(event, {
+          projectId: project.id,
+          projectName: project.name,
+          projectUrl: connectProjectUrl(project.id),
+          timelineId: req.timeline.id,
+          timelineName: req.timeline.name,
+          milestoneId: updated.id,
```
vikasrohit: @ngoctay @gondzo would it be better to send the `original` and `updated` milestone objects instead of sending specific fields of milestones? Rationale is that whoever is subscribing for these events, would be interested in knowing what were the previous and current values of milestone' status so that they can behave differently for different combinations of original vs updated values.

src/middlewares/userIdAuth.js
```diff
@@ -0,0 +1,77 @@
...
+    res.status(403).json(
+      util.wrapErrorResponse(req.id, 403, `Origin ${origin} is not allowed to access this authentication scheme`));
+    return res.end();
+  }
+
+  const userId = req.headers.authorization.split(bearerUserId)[1];
+
+  req.log.debug('Get m2m token');
+
+  m2m.getMachineToken(config.AUTH0_CLIENT_ID, config.AUTH0_CLIENT_SECRET)
```
vikasrohit: Could you please move it to `utils.js`? May be for now create a separate function than `getSystemUserToken` because that method is being used by some other api calls as well which might fail with m2m in absence of appropriate scope. So, for now a separate function would refactor the code to its desired place, and we can merge the two methods later as separate commit and validate its after effects.

src/events/busApi.js
ngoctay: Yeap, I will update to send the `original` and `updated` milestone

src/middlewares/userIdAuth.js
ngoctay: Do you want to make util function for:

1. Get m2m token? or
2. Get m2m token and make call to identity endpoint?

ngoctay:
@vikasrohit 
Sorry for the single complicated PR.

It's hard for me to separate it into single ones now.
When I was working on an issue, I usually found a bug in previous issues. Then I fixed it immediately in the commits for the current issue :(.

Will make separate branch and PR in the future.

src/util.js
```diff
@@ -297,13 +297,19 @@ _.assignIn(util, {
...
     httpClient.defaults.headers.common['Content-Type'] = 'application/json';
     httpClient.defaults.headers.common.Authorization = `Bearer ${jwtToken}`;
-    return httpClient.get(`${config.identityServiceEndpoint}users/${userId}`).then((response) => {
-      if (response.data && response.data.result
-        && response.data.result.status === 200 && response.data.result.content) {
-        return response.data.result.content;
-      }
-      return null;
-    });
+    return httpClient.get(`${config.identityServiceEndpoint}users`, {
```
vikasrohit: @ngoctay why we need this change? I think the old way of getting the details of a particular user should work as is.

src/middlewares/userIdAuth.js
vikasrohit: Helper method just to get the m2m token.

src/util.js
ngoctay: @vikasrohit 
I tried the old way, but it always returns 404. So I changed like that, and it works.

Could you please double check?

vikasrohit: @ngoctay I just checked it is working for me in dev. Please restore it to the old way, I would take care of it after merging if it does not work.

vikasrohit:
@ngoctay @gondzo Merging the PR into dev to test it out live.

PR #177, opened 2018-09-07
Some improvements for local deployment to work more out of the box.

maxceem:
- Improved example config `sample.local.js`
   - added `identityServiceEndpoint: "http://dockerhost:3001/"` otherwise we cannot create a new project, as this endpoint is being used to mock `getSystemUserToken` request which is required to call a member service, see https://github.com/topcoder-platform/tc-project-service/blob/dev/local/mock-services/authMiddleware.js
    - edited `memberServiceEndpoint: 'http://dockerhost:3001/v3/members',` to match the URL which is mocked by local deployment, see https://github.com/topcoder-platform/tc-project-service/blob/dev/local/mock-services/server.js#L24
     - edited `masterUrl: 'postgres://coder:mysecretpassword@dockerhost:5432/projectsdb',` as by default configuration Postgres is run on port `5432`, see https://github.com/topcoder-platform/tc-project-service/blob/dev/local/docker-compose.yml#L10

- Added one more user to the mocked member service. This user's JWT is provided in the https://github.com/topcoder-platform/tc-project-service/blob/dev/README.md. Otherwise if we create a new project using this JWT, we will get error, as memeber service doesn't know the user encoded in the token.

### DIFF
--- a/config/sample.local.js
+++ b/config/sample.local.js
@@ -5,6 +5,7 @@ if (process.env.NODE_ENV === 'test') {
   config = require('./test.json');
 } else {
   config = {
+    identityServiceEndpoint: "http://dockerhost:3001/",
     authSecret: 'secret',
     authDomain: 'topcoder-dev.com',
     logLevel: 'debug',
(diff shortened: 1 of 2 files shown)


vikasrohit:
LGTM

Issue #176, opened 2018-08-29
Add new field `lastActivityAt` in projects and update it from different child entity events


vikasrohit:
required to fix https://github.com/appirio-tech/connect-app/issues/2344

RishiRajSahu:
Done and pushed to dev.

PR #180, opened 2018-09-17
Feature/project activity

maxceem:
Implementation for https://github.com/topcoder-platform/tc-project-service/issues/176

It's based on the winning submission from the challenge https://www.topcoder.com/challenges/30070732

Some fixes/improvements have to be still made before merging.


migrations/20180910_project_activity.sql
```diff
@@ -0,0 +1,18 @@
...
+--   projects:
+--     added column `lastActivityAt`
+--     added column `lastActivityUserId`
+
+--
+-- projects
+
+-- Add new columns
+ALTER TABLE projects ADD COLUMN "lastActivityAt" timestamp with time zone;
+ALTER TABLE projects ADD COLUMN "lastActivityUserId" INT;
```
vikasrohit: @maxceem @RishiRajSahu we need to use string here because we are mostly using string in project models to capture the user as we might have situations where we have to store non number values like `coderbot` or `system`

maxceem: @vikasrohit oh, it's one of the questions I was interested to clarify. As inside project service, for user ids we use two types `INTEGER` and `BIGINT`. So probably we have to agree which one is better to use for the future. For example here inside project model we use two types for user ids https://github.com/topcoder-platform/tc-project-service/blob/dev/src/models/project.js#L41-L42

I cannot find any place where we use string for user ids though, could you please point to any place for reference?

vikasrohit: I think I was confused with message api where we used to have non numerical user ids because of discourse and we have to keep that during migration to allow system/coderbot posts.
I guess in project service we can keep BIGINT to be consistent. However, I do have one edge case where a topic/post is updated by a user like `coderbot` (it happens whenever coderbot posts a message e.g. on project creation it publishes a welcome message) or `system` (this is most probably not going to happen now because we are away from discourse now), what we would put in `lastActivityUserId` field in project service?

maxceem: @vikasrohit 

> I do have one edge case where a topic/post is updated by a user like coderbot

That's a very good catch.

@vikasrohit @RishiRajSahu 

Are we interested to update activity on the project which is done by a bot? The simple solution would be to check if `initiatorUserId` is a number or check if it's an existent user and only, in that case, to update activity. But it's only if we are not interested in activity by bots.

Otherwise I guess, we have to make `lastActivityUserId` as a string. So probably after we will show in the project list that latest activity in the project has been done by `Coder Bot`:

image

maxceem: @vikasrohit Added commit: use string for `lastActivityUserId`.

PR #190, opened 2018-10-05
Possible fix for admin reindex endpoint issue

vikasrohit:
fyi @RishiRajSahu


vikasrohit:
Merging because it passed the test step and failed only the deploy step because of AWS issue.

PR #192, opened 2018-10-24
Multi service deployment to avoid silent failure of kafka consumers

vikasrohit:
It also facilitates the better fault tolerance and better debugging in terms of separate cloud watch logs for api and consumers.


vikasrohit:
Closing the PR as we have rather implemented multi service deployment #192  for the same purpose because multi container task deployment was not serving the purpose.

PR #212, opened 2019-01-03
Feature/team management

gondzo:
@RishiRajSahu before deployment, the following migration should be executed `migrations/20181201_create_project_member_invites.sql`

### DIFF
--- a/src/events/projectMemberInvites/index.js
+++ b/src/events/projectMemberInvites/index.js
@@ -1,0 +1,76 @@
+/**
+ * Event handlers for project member invite create and update
+ */
+import _ from 'lodash';
+import Promise from 'bluebird';
+import { updateESPromise } from '../projectMembers';
+
+/**
+ * Project member invite careted event handler
+ * @param  {Object} logger  logger
+ * @param  {Object} msg     event payload
+ * @param  {Object} channel channel to ack / nack
+ * @return {undefined}
+ */
+const projectMemberInviteCreatedHandler = Promise.coroutine(function* a(logger, msg, channel) {
+  try {
+    const origRequestId = msg.properties.correlationId;
+    const newInvite = JSON.parse(msg.content.toString());
+    const projectId = newInvite.projectId;
+
+    // handle ES Update
+    // add new invite to document invites array
+    const updateDocPromise = Promise.coroutine(function* (doc) { // eslint-disable-line
+      // now merge the updated changes and reindex the document
+      const invites = _.isArray(doc._source.invites) ? doc._source.invites : []; // eslint-disable-line no-underscore-dangle
+      invites.push(newInvite);
+      return _.merge(doc._source, { invites }); // eslint-disable-line no-underscore-dangle
+    });
+
+    yield updateESPromise(logger, origRequestId, projectId, updateDocPromise);
+    logger.debug('elasticsearch index updated successfully');
+    channel.ack(msg);
+  } catch (error) {
+    logger.error('Error handling projectMemberInviteCreated Event', error);
+    // if the message has been redelivered dont attempt to reprocess it
+    channel.nack(msg, false, !msg.fields.redelivered);
+  }
+});
+
+/**
+ * Project member invite updated event handler
+ * @param  {Object} logger  logger
+ * @param  {Object} msg     event payload
+ * @param  {Object} channel channel to ack / nack
+ * @return {undefined}
+ */
+const projectMemberInviteUpdatedHandler = Promise.coroutine(function* a(logger, msg, channel) {
+  try {
+    const origRequestId = msg.properties.correlationId;
+    const updatedInvite = JSON.parse(msg.content.toString());
+    const projectId = updatedInvite.projectId;
+
+    // handle ES Update
+    // remove invite in document invites array, based on either userId or email
+    const updateDocPromise = Promise.coroutine(function* (doc) { // eslint-disable-line
+      // now merge the updated changes and reindex the document
+      const invites = _.isArray(doc._source.invites) ? doc._source.invites : []; // eslint-disable-line no-underscore-dangle
+      _.remove(invites, invite => (!!updatedInvite.email && invite.email === updatedInvite.email) ||
+          (!!updatedInvite.userId && invite.userId === updatedInvite.userId));
+      return _.merge(doc._source, { invites }); // eslint-disable-line no-underscore-dangle
+    });
+
+    yield updateESPromise(logger, origRequestId, projectId, updateDocPromise);
+    logger.debug('elasticsearch index updated successfully');
+    channel.ack(msg);
+  } catch (error) {
+    logger.error('Error handling projectMemberInviteCreated Event', error);
+    // if the message has been redelivered dont attempt to reprocess it
+    channel.nack(msg, false, !msg.fields.redelivered);
+  }
+});
+
+module.exports = {
+  projectMemberInviteCreatedHandler,
+  projectMemberInviteUpdatedHandler,
+};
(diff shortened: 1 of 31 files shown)


src/events/projectMembers/index.js
```diff
@@ -83,10 +83,13 @@ const projectMemberAddedHandler = Promise.coroutine(function* a(logger, msg, cha
...
       const memberDetails = yield util.getMemberDetailsByUserIds([newMember.userId], logger, origRequestId);
       const payload = _.merge(newMember, _.pick(memberDetails[0], 'handle', 'firstName', 'lastName', 'email'));
-      // now merge the updated changes and reindex the document
+      // now merge the updated changes and reindex the document for members
       const members = _.isArray(doc._source.members) ? doc._source.members : []; // eslint-disable-line no-underscore-dangle
       members.push(payload);
-      return _.merge(doc._source, { members }); // eslint-disable-line no-underscore-dangle
+      // now merge the updated changes and reindex the document for invites
+      const invites = _.isArray(doc._source.invites) ? doc._source.invites : []; // eslint-disable-line no-underscore-dangle
+      _.remove(invites, invite => invite === payload.email || invite === payload.userId);
```
RishiRajSahu: @gondzo ```invite.email === payload.email || invite.userId === payload.userId``` instead ?

src/events/projectMemberInvites/index.js
```diff
@@ -0,0 +1,76 @@
...
+    const origRequestId = msg.properties.correlationId;
+    const updatedInvite = JSON.parse(msg.content.toString());
+    const projectId = updatedInvite.projectId;
+
+    // handle ES Update
+    // remove invite in document invites array, based on either userId or email
+    const updateDocPromise = Promise.coroutine(function* (doc) { // eslint-disable-line
+      // now merge the updated changes and reindex the document
+      const invites = _.isArray(doc._source.invites) ? doc._source.invites : []; // eslint-disable-line no-underscore-dangle
+      _.remove(invites, invite => (!!updatedInvite.email && invite.email === updatedInvite.email) ||
```
RishiRajSahu: @gondzo this check is skipping updating to ES index when there is an update in status for already invited email or userID, isn't it needed to keep updated status of invite also re-indexed ?

src/routes/projects/get.js
```diff
@@ -64,6 +64,10 @@ module.exports = [
         if (attachments) {
           project.attachments = attachments;
         }
+        return models.ProjectMemberInvite.getPendingInvitesForProject(projectId);
+      })
+      .then((invites) => {
+        project.invites = invites;
```
RishiRajSahu: @gondzo not able to recall where this invites will be used on UI side?

src/routes/projectMembers/create.js
```diff
@@ -1,88 +1,72 @@
...
 import util from '../../util';
-import { PROJECT_MEMBER_ROLE, MANAGER_ROLES, EVENT } from '../../constants';
+import { USER_ROLE, PROJECT_MEMBER_ROLE, MANAGER_ROLES, INVITE_STATUS } from '../../constants';
+import models from '../../models';
 
 /**
  * API to add a project member.
- *
+ * add members directly (only managers and copilots)
+ * user being added is current user
```
vikasrohit: We might need the original functionality as well. I mean, we used this endpoint to add users to the team when other members of the project team are not able to add team members e.g. we recently faced this issue because of issue in member api. So, we added such members directly by calling this endpoint. Now as this endpoint is only supporting addition of the requesting user, we won't be able to add members to the project via API which might cause some troubles to support team. So, should we create a new endpoint to provide that functionality back? Or rather create new endpoint for self addition and keep the original one as is?

src/events/projectMembers/index.js
gondzo: good catch, will fix

src/routes/projectMembers/create.js
gondzo: the idea of team management is having users consent to join the project, so maybe we can just add a different endpoint for admin/support users? Though if we want that consent idea pushed all the way, we should also have a field where support can add a reason for adding a member manually (support case id, or similar)

src/events/projectMembers/index.js
gondzo: done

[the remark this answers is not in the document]
gondzo: done

src/routes/projectMemberInvites/create.js
```diff
@@ -0,0 +1,174 @@
...
+      }
+
+        // validate each userId is not already a member
+      const alreadyMembers = [];
+      _.forEach(members, (member) => {
+        if (invite.userIds.includes(member.userId)) {
+          alreadyMembers.push(member.userId);
+        }
+      });
+      if (alreadyMembers.length > 0) {
```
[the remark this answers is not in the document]
gondzo: done

src/routes/projectMemberInvites/create.js
```diff
@@ -0,0 +1,174 @@
...
+              _.forEach(invites, (i) => {
+                if (invite.userIds) {
+                  if (invite.userIds.includes(i.userId)) {
+                    alreadyInvites.push(i.userId);
+                  }
+                } else if (invite.emails.includes(i.email)) {
+                  alreadyInvites.push(i.email);
+                }
+              });
+              if (alreadyInvites.length > 0) {
```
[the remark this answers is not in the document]
gondzo: done

src/routes/projectMembers/create.js
vikasrohit: Yes, consent is also one of the goals, however, major goal was to ease the process of the joining a project. So, I guess we are good to have such end point where we can add member using those endpoints. And support team already logs the reason in the ticket where they are requested for adding a team member, so I guess we don't need to make efforts to introduce another field for tracking the reason.
Endpoint like previous version of this file would suffice, only improvement we can do there is to restrict that by administrator roles so that no one else can do that.

src/routes/projectMemberInvites/update.js
```diff
@@ -0,0 +1,119 @@
...
+        const err = new Error(
+          `invite not found for project id ${projectId}, email ${putInvite.email} and userId ${putInvite.userId}`,
+        );
+        err.status = 404;
+        return next(err);
+      }
+
+      req.log.debug('Chekcing user permission for updating invite');
+      let error = null;
+      if (putInvite.status === INVITE_STATUS.CANCELED) {
```
vikasrohit: I think there is some edge cases for these conditions. IMO, any authorized topcoder user can change status of an invite, which does not have `userId` but has `email`, to `ACCEPTED` or `REJECTED`.

gondzo:
```
if (putInvite.status === INVITE_STATUS.CANCELED) {
        if (!util.hasRoles(req, MANAGER_ROLES) && invite.role !== PROJECT_MEMBER_ROLE.CUSTOMER) {
          error = `Project members can cancel invites only for ${PROJECT_MEMBER_ROLE.CUSTOMER}`;
        }
      } else if (!!putInvite.userId && putInvite.userId !== req.authUser.userId) {
        error = 'Project members can only update invites for themselves';
      }
```
managers can cancel the invites, other users can just update their own invites

gondzo: you're right. fixed

vikasrohit: Great!! Could you please add a unit test for the same?

vikasrohit: Created issue for adding unit test https://github.com/topcoder-platform/tc-project-service/issues/214. Resolving the comment.

src/routes/projectMembers/create.js
RishiRajSahu: Created issue to keep track of this https://github.com/topcoder-platform/tc-project-service/issues/215.
Resolving.

PR #218, opened 2019-01-14
fix connect app issue 2741 - [Team Management 2.0] Handle the case when email entered to be invited is already topcoder user

maxceem:
Fix for Connect App issue https://github.com/appirio-tech/connect-app/issues/2741 - [Team Management 2.0] Handle the case when email entered to be invited is already topcoder user

### DIFF
--- a/src/routes/projectMemberInvites/create.js
+++ b/src/routes/projectMemberInvites/create.js
@@ -118,9 +118,35 @@ module.exports = [
           data.userId = null;
 
           if (invite.emails) {
+            // if for some emails there are already existent users, we will invite them by userId,
+            // to avoid sending them registration email
+            const existentUsers = util.lookupUserEmails(req, invite.emails).map((user) => {
+              const userWithNumberId = {};
+              _.assign(userWithNumberId, user, {
+                id: parseInt(user.id, 10),
+              });
+              return userWithNumberId;
+            });
+            // for existent users - invite by ids
+            const existentUserIds = _.map(existentUsers, 'id');
+            // the rest of email of non-existent users, so we will invite them by email
+            const nonExistentUserEmails = invite.emails.filter(inviteEmail =>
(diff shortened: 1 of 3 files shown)


src/routes/projectMemberInvites/create.spec.js
```diff
@@ -275,6 +277,57 @@ describe('Project Member Invite create', () => {
...
+        .expect(201)
+        .end((err, res) => {
+          if (err) {
+            done(err);
+          } else {
+            const resJson = res.body.result.content[0];
+            should.exist(resJson);
+            resJson.role.should.equal('customer');
+            resJson.projectId.should.equal(project2.id);
+            resJson.userId.should.equal(12345);
```
RishiRajSahu: @maxceem can we add check for email Id as well here ?

src/routes/projectMemberInvites/create.js
```diff
@@ -118,9 +118,35 @@ module.exports = [
...
+            const nonExistentUserEmails = invite.emails.filter(inviteEmail =>
+              !_.find(existentUsers, { email: inviteEmail }),
+            );
+
+            // remove invites for users that are invited already
+            _.remove(existentUserIds, userId => _.some(invites, i => i.userId === userId));
+            existentUserIds.forEach((userId) => {
+              const dataNew = _.clone(data);
+              _.assign(dataNew, {
+                userId,
```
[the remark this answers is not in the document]
maxceem: As we create a new object `const dataNew = _.clone(data);` and modify only `dataNew` means we don't modify `data`. So we don't need to set `data.userId = null;` after.

src/routes/projectMemberInvites/create.spec.js
maxceem: Added check email.

maxceem:
@RishiRajSahu added the check for email in unit tests. Lint is fine.

PR #247, opened 2019-02-09
Feature/attachment permissions

gondzo:
need to execute sql migration before deployment


swagger.yaml
```diff
@@ -2346,6 +2346,12 @@ definitions:
         type: number
         format: float
         description: The size of attachment
+      userIds:
```
vikasrohit: Do we need to rename this one to `allowedUsers`?

gondzo: yup. done.

Issue #245, opened 2019-02-07
Support filtering project list using substring for id, name, REF, customers and managers


maxceem:
To support the functionality described in this issue https://github.com/appirio-tech/connect-app/issues/2804 we have to support the ability to filter results of `GET /projects` endpoint using additional criteria.

#### Here is how the task looks from the point of view of Connect App:
- `ID`: User can type in a full ID or a portion (Ex. “55..) and all results matching that ID should be filtered.
- `Project name`: User can type in a full name or a portion (Ex. Eaton…) and all results matching that name should be filtered.
- `Reference Code`: User can type in a full code/tag or a portion (Ex. Zur…) and all results matching that name should be filtered. 
- `Customer`: Customer name should be filterable. User can type in a full name or a portion (Ex. Jim…) and all results matching that name should be filtered.
- `Managers`: Managers should be filterable. User can type in a name/handle and all results matching that should be filtered.

Implementation details should be added here as some of these filters cannot be done directly.

fyi @vikasrohit

maxceem:
As all the data we want to filter by is added to Elastic Search index we can filter by all of these properties.

maxceem:
As we already have filter by `id` which can be done by equality comparison `id=1234` where we compare the whole id. Or we can provide the list of id `id=in(1234,5678,9012)`.

For the new feature we can use syntax which would well matched with ES syntax:
- `id=12*` - find all projects with ids starts from `12`, `id=*12` - find all projects with id finishied with 12, `id=*12*` find all projects with id containing 12. While the current syntax `id=12` will find only project with id which equals to 12 as it's done now.

Same for filters for other properties.

vigneshTheDev:
> As all the data we want to filter by is added to Elastic Search index we can filter by all of these properties.

Looks like Customer/Manager names are not available in elastic search index. Should we query the db for these?

maxceem:
I thinkg they have to be in the ES, as per source code:
- on project creation: https://github.com/topcoder-platform/tc-project-service/blob/dev/src/events/projects/index.js#L38
- on member adding: https://github.com/topcoder-platform/tc-project-service/blob/dev/src/events/projectMembers/index.js#L85

Note, sometimes member API fails to respond with correct data and for some project we wouldn't have this data. But I guess locally you use mocked member API, so it should always work.

vikasrohit:
@maxceem Closing the issue. We would take care of including ID in search later.

PR #251, opened 2019-02-11
Add new filters for projects list

maxceem:
winning submission for challenge 30083089 - Topcoder Connect - Filter project table by columns, fix for issue #245

The quality of the submission looks great to me.

### DIFF
--- a/migrations/elasticsearch_sync.js
+++ b/migrations/elasticsearch_sync.js
@@ -209,7 +209,8 @@ function getRequestBody(indexName) {
         },
       },
       id: {
-        type: 'long',
+        type: 'string',
+        index: 'not_analyzed',
       },
       members: {
         type: 'nested',
--- a/postman.json
+++ b/postman.json
@@ -1486,6 +1486,120 @@
 					"response": []
 				},
 				{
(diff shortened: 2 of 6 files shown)


migrations/elasticsearch_sync.js
```diff
@@ -209,7 +209,8 @@ function getRequestBody(indexName) {
         },
       },
       id: {
-        type: 'long',
+        type: 'string',
+        index: 'not_analyzed',
```
maxceem: @vikasrohit this change has been done to the ES sync script which looks like required to filter by substring in `id`.

I don't see any drawbacks of this change. But maybe you have any thoughts on this.

maxceem:
Accidently, removed and after restored the branch.

migrations/elasticsearch_sync.js
RishiRajSahu: @maxceem can we separate out this id substring change from this PR if we have to reindex all projects again in order to make the filtering by ID substring enable ?

maxceem:
@RishiRajSahu I'm closing this PR as I've created another one https://github.com/topcoder-platform/tc-project-service/pull/256 which doesn't have an update for ID.

PR #254, opened 2019-02-19
Copilot workflow changes

gondzo:


src/events/busApi.js
```diff
@@ -721,5 +731,27 @@ module.exports = (app, logger) => {
       status,
       initiatorUserId: req.authUser.userId,
     }, logger);
+
+    if (role === PROJECT_MEMBER_ROLE.COPILOT && status === INVITE_STATUS.ACCEPTED) {
```
RishiRajSahu: @gondzo should we send only copilot events only in this case and bypass the event on line https://github.com/topcoder-platform/tc-project-service/pull/254/files#diff-42ae3999a1a28855656e72e3e1e1ea71R727 ? otherwise I think it would result in duplicate emails being sent from tc-notifications ?

src/events/busApi.js
```diff
@@ -707,9 +708,18 @@ module.exports = (app, logger) => {
       email,
       initiatorUserId: req.authUser.userId,
     }, logger);
+
+    if (role === PROJECT_MEMBER_ROLE.COPILOT) {
```
RishiRajSahu: @gondzo same as the issue mentioned in case of updated event below, should we bypass create event in case of requested event event thrown ?

vikasrohit: Yes, we should skip raising the created event in this case.

src/constants.js
```diff
@@ -130,7 +131,10 @@ export const BUS_API_EVENT = {
 
   // Project Member Invites
   PROJECT_MEMBER_INVITE_CREATED: 'notifications.connect.project.member.invite.created',
+  PROJECT_MEMBER_INVITE_REQUESTED: 'notifications.connect.project.member.invite.requested',
   PROJECT_MEMBER_INVITE_UPDATED: 'notifications.connect.project.member.invite.updated',
+  PROJECT_MEMBER_COPILOT_ADDED: 'notifications.connect.project.member.copilot.added',
```
vikasrohit: Renaming of events needed as suggested in other PR https://github.com/topcoder-platform/tc-notifications/pull/89

src/constants.js
```diff
@@ -130,7 +131,10 @@ export const BUS_API_EVENT = {
...
+  PROJECT_MEMBER_COPILOT_REFUSED: 'notifications.connect.project.member.copilot.refused',
```
vikasrohit: Renaming of events needed as suggested in other PR https://github.com/topcoder-platform/tc-notifications/pull/89

src/routes/projectMemberInvites/create.js
```diff
@@ -250,7 +250,7 @@ module.exports = [
                             { correlationId: req.id },
                         );
                     // send email invite (async)
-                    if (v.email && !v.userId) {
+                    if (v.email && !v.userId && v.role !== PROJECT_MEMBER_ROLE.COPILOT) {
```
vikasrohit: Can we better check the status (`v.status === 'pending'`) of the invite itself rather than depending of the role? We might have some future requirements where we might need to add more roles where we need workflow like this.

gondzo: no, we don't send the email for copilots only - invite status is still the same regardless of role

Issue #183, opened 2018-09-21
Removing usage of SystemUserToken 


RishiRajSahu:
With the incorporation of m2m, we have to remove the usage of` SystemUserToken` once the `https://api.topcoder-dev.com/v3/authorizations` endpoints supports m2m.

maxceem:
@RishiRajSahu I'm currently checking if we can fully move to M2M token.

Here are all the places/service where we currently use `SystemUserToken`:

| Code | Service |
|--|--|
| https://github.com/topcoder-platform/tc-project-service/blob/dev/src/events/projectMembers/index.js#L51 | Topcoder Direct |
| https://github.com/topcoder-platform/tc-project-service/blob/dev/src/events/projectMembers/index.js#L122 | Topcoder Direct |
| https://github.com/topcoder-platform/tc-project-service/blob/dev/src/util.js#L342 | Member Service |
| https://github.com/topcoder-platform/tc-project-service/blob/dev/src/util.js#L367 | Identity Service |
| https://github.com/topcoder-platform/tc-project-service/blob/dev/src/util.js#L452 | Identity Service |

Do you think these services `Topcoder Direct`, `Member Service` and `Identity Service` would accept a `token` obtained by M2M auth?

cc @vikasrohit

vikasrohit:
@maxceem I think we can get rid of system user token for member service endpoint and both of identity endpoints.
https://github.com/appirio-tech/ap-member-microservice/blob/dev/service/src/main/java/com/appirio/service/member/resources/MemberSearchResource.java#L90
https://github.com/appirio-tech/tc1-api-core/blob/dev/tech.core/tech.core.service.identity/src/main/java/com/appirio/tech/core/service/identity/resource/RoleResource.java#L103
https://github.com/appirio-tech/tc1-api-core/blob/dev/tech.core/tech.core.service.identity/src/main/java/com/appirio/tech/core/service/identity/resource/UserResource.java#L374

Only trouble is with direct api. So, we have to leave the method in place for calling direct api, however, we can definitely ignore, for local setup of participants, the usage of system user token for those calls as we don't have very tight integration with direct api yet and it just used to create a direct project and add members to that project occasionally.

vikasrohit:
Or may be if we can generate a PR for direct api to honour the m2m token, we can ask the owner of the repo to get it validated and merged.

maxceem:
[...]

PR #266, opened 2019-03-15
Use M2M token for Identity and Member services

maxceem:
- use M2M token for Identity and Member services
- improved README to cover two possible ways of running locally with and without M2M token provided

This PR almost fixes https://github.com/topcoder-platform/tc-project-service/issues/183


RishiRajSahu:
@maxceem Should we make the branch deployable and test these changes on dev before merging to dev, just to ensure that nothing breaks ?

maxceem:
@RishiRajSahu Sure, I'll re-create this PR using branch on the Topcoder repo, so it will be tested.

PR #268, opened 2019-03-15
use M2M token for Identity and Member services

maxceem:
- use M2M token for Identity and Member services
- improved README to cover two possible ways of running locally with and without M2M token provided

This PR almost fixes https://github.com/topcoder-platform/tc-project-service/issues/183


maxceem:
@RishiRajSahu this is tested now

PR #278, opened 2019-03-19
Github issue#277,Changes required for project member invite update endpoint

vikasrohit:
Potential fix


vikasrohit:
Yes, you are correct. But just being pedantic for that hypothetical case. Also, we may get new features where either there is similar workflow for non copilot users as well or very probable case is allowing copilot invitations by email.
So, this is just to be more logical in code because current code `userId: _.get(updatedInvite, 'userId', req.authUser.userId),` has potential(may be directly calling the API) of passing wrong `userId` to the project_member record.

maxceem:
Ah, ok. For a hypothetical case, this fix looks good for me.

PR #283, opened 2019-03-27
Feature/project templates refactoring

maxceem:
The work initially has been done in the CODE challenge https://www.topcoder.com/challenges/30086308/?type=develop but has various issues and didn't get the passing score.

It was fixed via private task https://www.topcoder.com/challenges/30087019

All the issues were well elaborated. In general the quality of submission looks good.


src/routes/timelines/create.js
```diff
@@ -18,7 +18,7 @@ const schema = {
   body: {
     param: Joi.object().keys({
       id: Joi.any().strip(),
-      name: Joi.string().max(255).required(),
+      name: Joi.string().max(45).required(),
```
vikasrohit: I think this change is unrelated to the PR.

migrations/20190317_refactor_project_templates.sql
```diff
@@ -0,0 +1,9 @@
+--
+-- project_templates
+--
+ALTER TABLE project_templates ALTER COLUMN "scope" DROP NOT NULL;
```
vikasrohit: I think we won't need this script for near future as we are not going to migrate all templates immediately. So, to be safe and unwanted execution of this script, should we remove the `drop` statements?

maxceem: Is it possible that you confused `DROP NOT NULL` command with `DROP COLUMN`? Here we only make the field optional not removing it.

If we don't make `scope` optional here than I'm not sure how we would create new `projectTemplates` without this field? As by a new way of creating projectTemplates we wouldn't define `scope` and `phases` and will define `form`, `priceConfig` and `planConfig` instead.

src/routes/timelines/create.js
maxceem: Yes, forgot to revert it. Removed.

migrations/20190316_extract_scope_from_project_templates.sql
```diff
@@ -0,0 +1,89 @@
+--
+-- form
+--
+
+CREATE TABLE form (
+    id bigint NOT NULL,
+    "key" character varying(45) NOT NULL,
+    "version" bigint DEFAULT 1 NOT NULL,
+    "revision" bigint DEFAULT 1 NOT NULL,
+    "scope" json DEFAULT '{}'::json NOT NULL,
```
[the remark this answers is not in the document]
maxceem: I didn't come up with better name. I think naming it `config` will make it clear. It would be same to other models and would fit any information we could add to this property later.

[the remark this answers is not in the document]
vikasrohit: Agreed.

migrations/20190317_refactor_project_templates.sql
vikasrohit: Yep, exactly. I misread that. 😬 Apologies..

vikasrohit:
@maxceem merging the PR as we are going to take care of renaming of the `scope` and `phases` field in a separate task.

maxceem:
@vikasrohit if you don't run migration scripts from this PR yet we can just rename these fields in migration scripts, rather than creating new migration scripts which would rename them to `config` in the DB. Or maybe we can revert this PR until this renaming is done. What do you think?

vikasrohit:
Yep, I haven't run the scripts yet. I was waiting for the PR to be deployed first. :) Yes, update the current migration script it self.

PR #285, opened 2019-04-01
DB query improvements 2

maxceem:
submission from challenge 30087679 - refactored project.searchText DB query


vikasrohit:
LGTM 👍

PR #290, opened 2019-04-12
Product templates use forms

maxceem:
winning submission from challenge 30088320 - Topcoder Project Service - Product templates use forms

Submission looks good to me.

I think to finish with this refactoring process we can do a couple of more things:
1. In this PR helper method has been created `util.checkModel` while in previous code this method has been defined several times: 
   - https://github.com/topcoder-platform/tc-project-service/blob/feature/product-templates-refactoring/src/routes/projectTemplates/create.js#L50
   - https://github.com/topcoder-platform/tc-project-service/blob/feature/product-templates-refactoring/src/routes/projectTemplates/update.js#L53
   - https://github.com/topcoder-platform/tc-project-service/blob/feature/product-templates-refactoring/src/routes/projectTemplates/upgrade.js#L55
   
   So we can reuse a new helper method in all these cases.

2. At the moment we have two ways of defining forms for both `ProductTemplates` and `ProjectTemplates`: old way inside the template and new way using Form model. 
    So both of fields are optional:

     image

    and
    
    image

    For `ProductTemplate` we should check in create/update endpoints that 
    - either `form` or `template` field is defined, but not both together.

    For `ProjectTemplate` we should check in create/update endpoints that:
    - either `form` or `scope` field must be defined, but not both together.
    -`form` or `priceConfig` cannot be defined together (they can be both `null` though)
    - either `phases` or `phaseConfig` field must be defined, but not both together.


vikasrohit:
> -form or priceConfig cannot be defined together (they can be both null though)

@maxceem I didn't get this statement. IMO, `form` and `priceConfig` would be set together in fact.

maxceem:
@vikasrohit right, it was a typo, it should be:
- `scope` or `priceConfig` cannot be defined together (they can be both `null` though)

I've updated in the comment above.

The main idea that we should validate that we define form template either new or old way, but not both together.

vikasrohit:
Okay, got it. Please feel free to run task for the pending items as you listed above, they look good to me.

Issue #299, opened 2019-05-02
New endpoints for fetching project phases and products from database


vikasrohit:
Currently we have list endpoint for fetching a project's phases/products which fetches the data from ES cluster. There are incidents where we are not able to get the phases/products in time for a project just after creating a project. So get rid of that situation, we need to have an endpoint which returns data from database without waiting for ES indexing.
Purpose of this task is to replicate the behaviour of current list endpoints for `projectPhases` and `phaseProducts` and use database instead of ES index to fetch the data from.

maxceem:
@vikasrohit to confirm.

Currently, we have endpoints 
- to list phases `GET /v4/projects/:projectId(\\d+)/phases` https://github.com/topcoder-platform/tc-project-service/blob/dev/src/routes/phases/list.js
- and products `GET /v4/projects/:projectId(\\d+)/phases/:phaseId(\\d+)/products` https://github.com/topcoder-platform/tc-project-service/blob/dev/src/routes/phaseProducts/list.js

We should implement endpoints to get data from DB for them
- `GET /v4/projects/:projectId(\\d+)/phases/db`
- `GET /v4/projects/:projectId(\\d+)/phases/:phaseId(\\d+)/products/db`

This should be done a similar way as now its done for projects:
- https://github.com/topcoder-platform/tc-project-service/blob/dev/src/routes/projects/list-db.js

vikasrohit:
Yes.

maxceem:
@vikasrohit just in case, this is already done and merged to DEV branch.

PR #300, opened 2019-05-03
DB endpoints

maxceem:
Done issue https://github.com/topcoder-platform/tc-project-service/issues/299

winning submission from challenge 30089841 - Topcoder Project Service - Create DB endpoints

### DIFF
--- a/src/models/projectPhase.js
+++ b/src/models/projectPhase.js
@@ -1,5 +1,3 @@
-/* eslint-disable valid-jsdoc */
-
 import _ from 'lodash';
 
 module.exports = function defineProjectPhase(sequelize, DataTypes) {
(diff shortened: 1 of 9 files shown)


src/models/projectPhase.js
```diff
@@ -44,62 +42,32 @@ module.exports = function defineProjectPhase(sequelize, DataTypes) {
...
+       * @param {Object} parameters the parameters
+       *          - sortField: the field that will be references when sorting
+       *          - sortType: ASC or DESC
+       *          - fields: the fields to retrieved
+       *          - projectId: the id of project
+       * @param {Object} log the request log
+       * @return {Object} the result rows and count
        */
-      searchText(parameters, log) {
-        // special handling for keyword filter
```
vikasrohit: @maxceem do you know why we need to remove this code block for these changes? I mean this endpoint might have been used at other places as well.

maxceem: @vikasrohit as we didn't use this method, I asked to remove it as an obsolete code. Also, I don't know about possible applications of this method, as basically we filter phases across multiple projects, but it looks like we only need to filter phases inside one project.

Let me know if we could possibly need this method and I would revert it.

vikasrohit: Okay. Makes sense. I am aware of the usage of this method either. Lets leave it removed.

PR #302, opened 2019-05-08
Implement dot notation email validation for gmail

mfikria:


mfikria:
@maxceem i've update the PR. Please review it again thanks!

mfikria:
@maxceem updated with:
1. change regex validation
2. update config for `compareEmail` function

maxceem:
Noticed one more thing:

- If we have an invitation with email `test@GMAIL.com` and try to invite email `test@gmail.com` - these emails would be treated differently.

The reason for this is that the method `compareEmail` is not symmetrical: 
- `compareEmail(email1, email2) !== compareEmail(email2, email1)` - sometimes

mfikria:
@maxceem i've updated the function:
1. fix regex for additional dot with '\.?' instead of '.?'
2. only allow gmail.com and googlemail.com as email domains
3. make email domain case insensitive

PR #304, opened 2019-05-11
issue 2969 fix

PrakashDurlabhji:


PrakashDurlabhji:
@maxceem status check is applied on "entitytoupate", which means if already it is completed in db then do not allow.

But if currentstatus is going to make it completed then allow it.

maxceem:
@PrakashDurlabhji hm, are you sure, this is how it looks to me:

image

PrakashDurlabhji:
@maxceem pr updated

PR #311, opened 2019-06-05
set SSO flag for invite events

gondzo:
SSO_REFCODES environment variable should be populated with ref codes for orgs that have SSO integration

FYI @RishiRajSahu


RishiRajSahu:
LGTM

PR #321, opened 2019-06-25
improve milestone hooks and change `referenceId` in milestone to BIGINT

mfikria:


mfikria:
@maxceem I cannot reproduce the issue since I've got success response and `timeline`, `milestone`, and `statusHistory` are created correctly in the DB. My steps to check the issue:
0. Kafka, rabbitmq, postgres, etc are run
1. npm i (with node 8.9)
2. npm run sync:db
3. npm run sync:es
4. CONNECT_USER_TOKEN=yourtoken npm run demo-data
5. Test POST /timelines with postman

maxceem:
For reference, here was the task:

#### Issue:
When we use bulkCreate method to create milestones the hook afterCreate is not called which leads to a couple of issues:
- when milestones are created in a bulk using POST /timelines endpoint, the StatusHistory records are not created. Also, statusHistory records are not populated by the POST /timelines endpoint even if they would be created, as afterCreate hook is not called, thus mapWithStatusHistory is not called either.
- I saw that you've updated unit tests in your submission and create statusHistory records manually, as in unit tests we also use bulkCreate which doesn't trigger afterCreate hook

#### Task:
1. fix model StatusHistory: referenceId should be BIGINT instead of STRING. Note, that migration script already uses bigint, so only the model should be fixed and any place in the code which uses string for referenceId instead of an integer like this one https://github.com/topcoder-platform/tc-project-service/blob/feature/milestone-pausing/src/models/milestone.js#L15. Update unit tests and swagger if they use string instead of bigint.
2. add afterBulkCreate hook which would create statusHistory records. Preferable, create statusHistory records in a bulk in this method if possible. Also, populate status history in this hook. 
3. Improve method mapWithStatusHistory to support array of milestones. So instead of querying DB several times for each milestone, we query DB once, find statusHistory records for all milestones, and after populate each milestone with statusHistory records which belong to that milestone. This would be useful inside afterBulkCreate hook and also inside afterFind hook in case when the "milestone" is an array.
4. Update unit tests: don't create StatusHisotry records manually as you've implemented in your last submission. After the fixes above they should be created automatically.

Issue #294, opened 2019-04-19
Add paused status for the milestones


vikasrohit:
Continuing from https://github.com/appirio-tech/connect-app/issues/2968

This issue is for handling the backend changes required for the new status. We would need to keep history of the milestone status changes so that we can enable caller of the API to calculate the time spent in the each of the state of the milestone and move the status back to previous status when user wants to resume the milestone.

fyi @maxceem @RishiRajSahu

maxceem:
@vikasrohit there is a couple of ways to go, would be great to have your view on them.

Does it make sense to make this new model universal so it could be reused by any other model not only the milestones in the future?
   1. So we can refer to the rows in this model by the `milestoneId`. In this case we can use it only for milestones. And if later we want to have the same functionality for other model we would need to create another model for that. 

      We can call model `MilestonePause`
       ```
       - id: Number, primary, autoincerement
       - milestoneId: Number (required)
       - startTime: Datatime (required)
       - endTime: Datetime
       - comment: Text (required)
       - createdBy
       - createdAt (will duplicate startTime)
       - updatedBy
       - updatedAt (will duplicate endTime)      
       ``` 
   2. Alternatively, the tracking model can refer other entities by two fields: `reference` and `referenceId`. So for the milestones `reference` would be always `milestones` and `referenceId` would be equal to the particular `milestoneId`. In such case we could reuse our tracking model for any other model by defining new `reference` values. As a disadvantage, theoretically, this tracking table could grow big if tracks pauses for too many models. Also, if we are not going to use it for any other models, this is not a necessary abstraction.

       We can call model `Pause`
       ```
       - id: Number, primary, autoincerement
       - reference: String (required)
       - referenceId: String (required)
       - startTime: Datatime (required)
       - endTime: Datetime
       - comment: Text (required)
       - createdBy
       - createdAt (will duplicate startTime)
       - updatedBy 
       - updatedAt (will duplicate endTime)
       ``` 
3. I guess we should make restictions for this model. 
    1. We can only define `startTime` when we create the records, and we cannot update `startTime`.
    2. We can only define `endTime` once. If it's already defined we cannot update it. 
    3. We can update the `comment` though.

4. If we make the restrictions above for all the users, then `createdAt` will always duplicate `startTime`. And `updatedAt` will always duplicate `endTime`. Unless admins can make changes to this model bypassing restrictions above.

vikasrohit:
[...]

PR #322, opened 2019-06-28
Feature/milestone pausing

maxceem:
Implementation for https://github.com/topcoder-platform/tc-project-service/issues/294


src/routes/timelines/create.js
```diff
@@ -51,9 +51,9 @@ module.exports = [
 
     let result;
     // Save to DB
-    models.sequelize.transaction(() => {
+    models.sequelize.transaction((tx) => {
```
vikasrohit: With https://github.com/topcoder-platform/tc-project-service/blob/02d71977909ede3bebce8ce91a9072fbdc7ea01f/src/models/index.js#L5, Isn't it automatically passing the transaction everywhere we need it?

maxceem: That's an interesting catch @vikasrohit. I wasn't aware that we were using `cls`. As per my testings, it feels that transactions didn't work (partials changes weren't reverted in case something failed).  Though it might be some other reasons as I could make several fixes at once as I was sure that it wouldn't work without directly defining transactions.

I've forced adding transactions in several places in this PR 
image

Not sure what is better. Merge this PR and log issue to investigate whether transactions work. Or investigate first and adjust the PR is necessary before merging.

Which way would you prefer?

vikasrohit: Merge this PR and log issue to investigate whether transactions work.

maxceem: Logged the issue https://github.com/topcoder-platform/tc-project-service/issues/323

maxceem:
@vikasrohit forgot to mention that migration script should be run for this PR `migrations/20190502_status_history_create.sql`.

vikasrohit:
> forgot to mention that migration script should be run for this PR migrations/20190502_status_history_create.sql.

Done.

PR #330, opened 2019-07-04
fixed #234 prevent user invite if already added

### LINKED ISSUE #234: [$120] Don't let invite user by email if he is already in the team or already invited
### Actual behaviour
Example, I'm already is a Manager in **Topcoder team** of this project https://connect.topcoder-dev.com/projects/7709

I invite myself to the **Project Team** using email my address and server accepts my requests and creates an invitation.
image

After I reload the page I got invitation popup even though I'm already in the Topcoder team.

image

If I click **Join Project** I would get error that I'm already in the team.

### Expected behaviour
When we invite user by email which is already in the Topcoder or Project Team, server should reject such request, don't create invitation and return an error, that a user with such an email is already in the team.

This issue is somehow connected with https://github.com/topcoder-platform/tc-project-service/issues/233

fyi @RishiRajSahu

romitgit:


romitgit:
@maxceem In the case of inviting a user by email when it is already a member of the project, wouldn't it be right to assume that he will already have a userId and in that case we invite them by userIDs only. And we are already throwing error for this case i.e. when allready added user in invited using userhandle so that will fix the email problem. 

Please clarify.

Thanks

maxceem:
@romitgit
> @maxceem In the case of inviting a user by email when it is already a member of the project, wouldn't it be right to assume that he will already have a userId and in that case we invite them by userIDs only. 

We can still technically invite users by email, even they are already registered on Topcoder and member of the project. Actually, we shouldn't invite users who is already a team member that's why we want to return an error in such case.

We should return an error the same way as a user has been invited, so the client-side can show the error message properly:
- if user is invited by `email` we return `email` in the error response
- if user has been invited by `handle` we return `userId` in the error response

Let me know if it helps, or you have any other questions.

maxceem:
@romitgit do you think you would have time to finish this issue shortly?

Btw, there is some new code in `cf18` branch, don't forget to pull it before continuing working on this issue.

PR #335, opened 2019-07-05
Add new tests for project update

### LINKED ISSUE #314: [$35] Prevent non admins to update the template id of a project
As of now a customer can update the template id of their projects by directly calling the API endpoint. Although, it does not do any harm but it can result in issues in rendering a project because data for the project would not be stored in the formatted as per the updated templated id.
So, we should prevent updating template id of an existing project for non admin users.

fyi @maxceem @RishiRajSahu

r0hit-gupta:
Fixes #314

### DIFF
--- a/src/routes/projects/update.spec.js
+++ b/src/routes/projects/update.spec.js
@@ -122,6 +122,12 @@ describe('Project', () => {
             userId: 40051332,
             createdBy: 1,
             updatedBy: 1,
+          }, {
+            projectId: project1.id,
+            role: 'member',
+            userId: 40051331,
+            createdBy: 1,
+            updatedBy: 1,
           }]);
         }).then(() => done());
     });
@@ -190,7 +196,7 @@ describe('Project', () => {
         })
         .send({
           param: {
-            templateId: '2',
+            templateId: 2,
           },
         })
         .expect('Content-Type', /json/)
@@ -202,7 +208,7 @@ describe('Project', () => {
             const result = res.body.result;
             result.success.should.be.true;
             result.status.should.equal(200);
-            result.content.templateId.should.not.equal('2');
+            result.content.templateId.should.equal(project1.templateId);
             done();
           }
         });
@@ -216,7 +222,7 @@ describe('Project', () => {
         })
         .send({
           param: {
-            templateId: '2',
+            templateId: 2,
           },
         })
         .expect('Content-Type', /json/)
@@ -228,7 +234,7 @@ describe('Project', () => {
             const result = res.body.result;
             result.success.should.be.true;
             result.status.should.equal(200);
-            result.content.templateId.should.not.equal('2');
+            result.content.templateId.should.equal(project1.templateId);
             done();
           }
         });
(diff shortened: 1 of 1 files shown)


src/routes/projects/update.spec.js
```diff
@@ -179,6 +182,84 @@ describe('Project', () => {
...
 
+    it('should not update the templateId for admin', (done) => {
+      request(server)
+        .patch(`/v4/projects/${project1.id}`)
+        .set({
+          Authorization: `Bearer ${testUtil.jwts.admin}`,
+        })
+        .send({
+          param: {
+            templateId: '2',
```
maxceem: `templateId` has type `BIGING`, so we should use a number, to make sure there is no other error, which prevents us from updating it.

src/routes/projects/update.spec.js
```diff
@@ -179,6 +182,84 @@ describe('Project', () => {
...
+
+    it('should not update the templateId for manager', (done) => {
+      request(server)
+        .patch(`/v4/projects/${project1.id}`)
+        .set({
+          Authorization: `Bearer ${testUtil.jwts.manager}`,
+        })
+        .send({
+          param: {
+            templateId: '2',
```
maxceem: `templateId` has type `BIGING`, so we should use a number, to make sure there is no other error, which prevents us from updating it.

src/routes/projects/update.spec.js
```diff
@@ -179,6 +182,84 @@ describe('Project', () => {
...
+
+    it('should not update the templateId for user', (done) => {
+      request(server)
+        .patch(`/v4/projects/${project1.id}`)
+        .set({
+          Authorization: `Bearer ${testUtil.jwts.member}`,
+        })
+        .send({
+          param: {
+            templateId: '2',
```
maxceem: `templateId` has type `BIGING`, so we should use a number, to make sure there is no other error, which prevents us from updating it.

src/routes/projects/update.spec.js
```diff
@@ -179,6 +182,84 @@ describe('Project', () => {
...
+        .expect('Content-Type', /json/)
+        .expect(200)
+        .end((err, res) => {
+          if (err) {
+            done(err);
+          } else {
+            const result = res.body.result;
+            result.success.should.be.true;
+            result.status.should.equal(200);
+            result.content.templateId.should.not.equal('2');
```
maxceem: For the assertion, it's better to check that the original value of `templateId` is still there, as it's important for us to keep inital value, just in case the value has been changed to some other.

src/routes/projects/update.spec.js
```diff
@@ -179,6 +182,84 @@ describe('Project', () => {
...
```
maxceem: For the assertion, it's better to check that the original value of `templateId` is still there, as it's important for us to keep inital value, just in case the value has been changed to some other.

src/routes/projects/update.spec.js
```diff
@@ -179,6 +182,84 @@ describe('Project', () => {
...
+        .expect('Content-Type', /json/)
+        .expect(403)
+        .end((err, res) => {
+          if (err) {
+            done(err);
+          } else {
+            const result = res.body.result;
+            result.success.should.be.false;
+            result.status.should.equal(403);
+            result.content.message.should.equal('You do not have permissions to perform this action');
```
maxceem: A normal user who is a member of the project can update project. So we should use a user who is a member of a project for testing and make sure that he can update the project but he cannot update `teamplteId`.

r0hit-gupta:
Fixed. Please review.

vikasrohit:
@maxceem I noticed that the PR was against dev, was it intentional?

maxceem:
Oh no, @vikasrohit reverted https://github.com/topcoder-platform/tc-project-service/pull/338

Will merge to `cf18` instead.

PR #336, opened 2019-07-06
Fix copilotAndAbove permission to check that users is a member of the project

tc-mfikria:


mfikria:
@maxceem ok, got it. I've updated the logic. Thanks!

PR #341, opened 2019-07-11
fixed topcoder-platform#234 : added checks for user already member of the team

romitgit:


romitgit:
@maxceem I have made the changes and here is the pull request https://github.com/topcoder-platform/tc-project-service/pull/341

src/routes/projectMemberInvites/create.js
```diff
@@ -63,11 +63,20 @@ const compareEmail = (email1, email2, options = { UNIQUE_GMAIL_VALIDATION: false
  *
  * @returns {Promise<Promise[]>} list of promises
  */
-const buildCreateInvitePromises = (req, invite, invites, data, failed) => {
+const buildCreateInvitePromises = (req, invite, invites, data, failed, members) => {
   const invitePromises = [];
   if (invite.userIds) {
     // remove invites for users that are invited already
-    _.remove(invite.userIds, u => _.some(invites, i => i.userId === u));
+    const errMessageForAlreadyInvitedUsers = 'User with such email is already invited to this project.';
```
maxceem: As invitation here uses `invite.userIds` so we have to use message with work `handle`.

src/routes/projectMemberInvites/create.js
```diff
@@ -63,11 +63,20 @@ const compareEmail = (email1, email2, options = { UNIQUE_GMAIL_VALIDATION: false
...
+    _.remove(invite.userIds, u => _.some(invites, (i) => {
+      if (i.userId === u) {
+        failed.push(_.assign({}, {
+          email: i.email,
```
maxceem: As invitation here uses `invite.userIds` so we have to return `userId` instead of `email`.

src/routes/projectMemberInvites/create.js
```diff
@@ -63,11 +63,20 @@ const compareEmail = (email1, email2, options = { UNIQUE_GMAIL_VALIDATION: false
...
+          message: errMessageForAlreadyInvitedUsers,
+        }));
+      }
+      return i.userId === u;
```
maxceem: In this code we check the same condition twice:
```js
if (i.userId === u) {
```

and 
```js
return i.userId === u;
```

Can we just check it once and inside the `if () {...}`?
```js
return true
```

src/routes/projectMemberInvites/create.js
```diff
@@ -96,8 +105,32 @@ const buildCreateInvitePromises = (req, invite, invites, data, failed) => {
...
+        const errMessageForAlreadyMemberUsers = 'User with such email is already a member of this project.';
+
+        _.remove(existentUsersWithNumberId, user => _.some(members, (m) => {
+          if (m.userId === Number(user.id)) {
+            failed.push(_.assign({}, {
+              email: user.email,
+              message: errMessageForAlreadyMemberUsers,
+            }));
+          }
+          return m.userId === user.id;
```
maxceem: In this code we check the same condition twice:

```js
if (m.userId === Number(user.id)) {
```

and 
```js
return m.userId === user.id;
```

Can we just check it once and inside the `if () {...}`?

```js 
return true
```

Note, that in your code two conditions suppose to be the same, but they are different. If we keep only one condition inside `if` and `return true` such issue wouldn't happen.

src/routes/projectMemberInvites/create.js
```diff
@@ -96,8 +105,32 @@ const buildCreateInvitePromises = (req, invite, invites, data, failed) => {
...
+        const errMessageForAlreadyInvitedUsers = 'User with such email is already invited to this project.';
+
+        _.remove(existentUsersWithNumberId, user => _.some(invites, (i) => {
+          if (i.userId === Number(user.id)) {
+            failed.push(_.assign({}, {
+              email: i.email,
+              message: errMessageForAlreadyInvitedUsers,
+            }));
+          }
+          return i.userId === user.id;
```
maxceem: In this code we check the same condition twice:

```js
if (m.userId === Number(user.id)) {
```

and 
```js
return m.userId === user.id;
```

Can we just check it once and inside the `if () {...}`?

```js 
return true
```

Note, that in your code two conditions suppose to be the same, but they are different. If we keep only one condition inside `if` and `return true` such issue wouldn't happen.

src/routes/projectMemberInvites/create.js
```diff
@@ -204,9 +248,18 @@ module.exports = [
...
       // remove members already in the team
-      _.remove(invite.userIds, u => _.some(members, m => m.userId === u));
+      _.remove(invite.userIds, u => _.some(members, (m) => {
+        if (m.userId === u) {
+          failed.push(_.assign({}, {
+            userId: m.userId,
+            message: errorMessageForAlreadyMemberUser,
+          }));
+        }
+        return m.userId === u;
```
maxceem: Same like a few palces above, it would be better to check condition only once and `return true` inside `if` to avoid issues.

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -480,14 +485,14 @@ describe('Project Member Invite create', () => {
...
+        .expect(403)
         .end((err, res) => {
           if (err) {
             done(err);
           } else {
-            const resJson = res.body.result.content.success;
+            const resJson = res.body.result.content.failed;
             should.exist(resJson);
-            resJson.length.should.equal(0);
+            resJson.length.should.equal(1);
```
maxceem: Could you please, also check, that faild object is correct? So it has the correct error message, and it return correct `userId`.

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -696,20 +701,20 @@ describe('Project Member Invite create', () => {
...
```
maxceem: Could you please, also check, that faild object is correct? So it has the correct error message, and it return correct `email`.

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -722,20 +727,20 @@ describe('Project Member Invite create', () => {
...
```
maxceem: Could you please, also check, that faild object is correct? So it has the correct error message, and it return correct `email`.

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -749,20 +754,20 @@ describe('Project Member Invite create', () => {
...
+          .expect(403)
           .end((err, res) => {
             if (err) {
               done(err);
             } else {
-              const resJson = res.body.result.content.success;
+              const resJson = res.body.result.content.failed;
               should.exist(resJson);
-              resJson.length.should.equal(0);
+              resJson.length.should.equal(1);
```
maxceem: Could you please, also check, that faild object is correct? So it has the correct error message, and it return correct `email`.

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -776,14 +781,14 @@ describe('Project Member Invite create', () => {
...
```
maxceem: Could you please, also check, that faild object is correct? So it has the correct error message, and it return correct `email`.

romitgit:
@maxceem thanks. Will make the changes asap.

romitgit:
@maxceem made some changes. Please review them

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -161,56 +184,56 @@ describe('Project Member Invite create', () => {
...
-        .end((err, res) => {
-          if (err) {
-            done(err);
-          } else {
-            res.body.result.status.should.equal(400);
-            done();
-          }
-        });
-        });
+    xit('should return 400 if neither userIds or email is presented',
```
maxceem: Why this test is disabled?

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -161,56 +184,56 @@ describe('Project Member Invite create', () => {
...
+              done(err);
+            } else {
+              res.body.result.status.should.equal(400);
+              done();
+            }
+          });
+      });
 
-    it('should return 403 if try to create copilot without MANAGER_ROLES', (done) => {
+    xit('should return 403 if try to create copilot without MANAGER_ROLES', (done) => {
```
maxceem: Why this test is disabled?

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -254,7 +277,7 @@ describe('Project Member Invite create', () => {
         });
     });
 
-    it('should return 403 if try to create copilot with MEMBER', (done) => {
+    xit('should return 403 if try to create copilot with MEMBER', (done) => {
```
maxceem: Why this test is disabled?

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -298,7 +321,7 @@ describe('Project Member Invite create', () => {
         });
     });
 
-    it('should return 201 and add new email invite as customer', (done) => {
+    xit('should return 201 and add new email invite as customer', (done) => {
```
maxceem: Why this test is disabled?

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -346,7 +369,7 @@ describe('Project Member Invite create', () => {
         });
     });
 
-    it('should return 201 and add new userId invite as customer for existent user when invite by email', (done) => {
+    xit('should return 201 and add new userId invite as customer for existent user when invite by email', (done) => {
```
maxceem: Why this test is disabled?

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -400,7 +423,7 @@ describe('Project Member Invite create', () => {
         });
     });
 
-    it('should return 201 and add new user invite as customer', (done) => {
+    xit('should return 201 and add new user invite as customer', (done) => {
```
maxceem: Why this test is disabled?

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -448,7 +471,101 @@ describe('Project Member Invite create', () => {
...
+        })
+        .expect('Content-Type', /json/)
+        .expect(403)
+        .end((err, res) => {
+          if (err) {
+            done(err);
+          } else {
+            const resJson = res.body.result.content.failed;
+            should.exist(resJson);
+            resJson[0].userId.should.equal(40051331);
```
maxceem: Can we also check, that the error message returned by the server is correct?

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -448,7 +471,101 @@ describe('Project Member Invite create', () => {
...
+        })
+        .expect('Content-Type', /json/)
+        .expect(403)
+        .end((err, res) => {
+          if (err) {
+            done(err);
+          } else {
+            const resJson = res.body.result.content.failed;
+            should.exist(resJson);
+            resJson[0].email.should.equal('romit.choudhary@rivigo.com');
```
maxceem: Can we also check, that the error message returned by the server is correct?

src/routes/projectMemberInvites/create.spec.js
```diff
@@ -448,7 +471,101 @@ describe('Project Member Invite create', () => {
...
+            resJson.length.should.equal(1);
+            server.services.pubsub.publish.neverCalledWith('project.member.invite.created').should.be.true;
+            done();
+          }
+        });
+    });
+
+    xit('should return 403 and empty response when trying add already invited member', (done) => {
```
maxceem: Why this test is disabled?

src/routes/projectMemberInvites/create.spec.js
maxceem: Can we also check, that the error message returned by the server is correct?

src/routes/projectMemberInvites/create.spec.js
maxceem: Can we also check, that the error message returned by the server is correct?

src/routes/projectMemberInvites/create.js
```diff
@@ -60,14 +60,25 @@ const compareEmail = (email1, email2, options = { UNIQUE_GMAIL_VALIDATION: false
  * @param {Array}  invites existent invites from DB
  * @param {Object} data    template for new invites to be put in DB
  * @param {Array}  failed  failed invites error message
+ * @param {Array} members  already members of the group
```
maxceem: I think it should be `project` instead of `group`.

romitgit:
Thanks @maxceem. The tests were disabled only for testing. Will enable them.
Also, will add error message check as well. 

One thing, is this still part of topcoder challenge?

maxceem:
> One thing, is this still part of topcoder challenge?

This is already not a part of the challenge, as the challenge has been finished some time ago. I would still get paid for this task if you complete it.

> Thanks @maxceem. The tests were disabled only for testing. Will enable them.
Also, will add error message check as well.

Please, don't disable tests as we should also check they existent tests still pass. I cannot test the submission if some tests are disabled.

romitgit:
@maxceem made the said changes. Please review.

maxceem:
Thank you @romitgit I'll review it tomorrow morning.

maxceem:
Sorry for the delay @romitgit. Today we are releasing the new version of Connect. Will for sure review your update tomorrow.

romitgit:
@maxceem No worries. Thanks for the update and good luck for the release.

romitgit:
@maxceem Done with review changes. Please review

Issue #315, opened 2019-06-11
New data models for work management


vikasrohit:
We need to introduce new data models and need to update existing ones to achieve the requirements/features needed for the new work management strategy. Though, we don't have final designs of all input and display requirements, we do have the feature details and we can determine common fields out of those details. So, we can start creating the model and API endpoints around it to be ahead of time and later update the end points when we have more detailed UI designs.

Technically, there is one new data model called `WorkStream` that is going to be introduced and it would have schema like:

```
id bigint,
name varchar(255),
type varchar(45) or enum (dev, qa, design, deploy),
status enum (draft, reviewed, active, paused, completed), // we might not need this in final version
projectId bigint,
createdAt timestamp,
updatedAt timestamp,
createdBy bigint,
updatedBy bigint,
deletedAt timestamp,
deletedBy bigint
```
New endpoints:
```
GET /projects/{projectId}/workstreams (fetch directly from database)
GET /projects/{projectId}/workstreams/{id} (fetch directly from database)
POST  /projects/{projectId}/workstreams (only admins should be able to directly create the workstreams)
PATCH  /projects/{projectId}/workstreams/{id} (it requires complex logic of authorization depending on the project template, for some templates we allow customers to be able to update the details while for others we don't)
DELETE  /projects/{projectId}/workstreams/{id} (only admins should be able to directly deleted the workstreams)
```

We need to keep one to many relationship (via cross ref table) between `WorkStream` and `Phase` object so that one work stream can have multiple phases. `Phase` object would remain mostly as is. For old projects a phase won't be associated with any work stream, so it is not mandatory for a phase to have workstream associate with it. After we have this relationship, following new endpoints are needed:

```
GET /projects/{id}/workstreams/{id}/works (we are calling a phase as work in new terminology without changing it in our data model. fetch phases from the database directly.)
GET /projects/{id}/workstreams/{id}/works/{id} (fetch directly from database)
POST  /projects/{id}/workstreams/{id}/works (it requires complex logic of authorization depending on the project template, for some templates we allow customers to be able to create work for a worksstream while for others we don't)
PATCH  /projects/{id}/workstreams/{id}/works/{id} (it requires complex logic of authorization depending on the project template, for some templates we allow customers to be able to update the details of a work while for others we don't)
DELETE  /projects/{id}/workstreams/{id}/works/{id} (it requires complex logic of authorization depending on the project template, for some templates we allow customers to be able to delete a work while for others we don't)
```

[...]

PR #342, opened 2019-07-11
Workstreams

maxceem:
Implementation for https://github.com/topcoder-platform/tc-project-service/issues/315

It has been accomplished by challenge https://www.topcoder.com/challenges/30093850/?type=develop

With some final-fixes included.


swagger.yaml
```diff
@@ -604,6 +604,440 @@ paths:
...
+          description: >
+            sort project phases by startDate, endDate, status, order. Default is
+            startDate asc
+          in: query
+          type: string
+      responses:
+        '200':
+          description: A list of project works
+          schema:
+            $ref: '#/definitions/ProjectPhaseListResponse'
```
vikasrohit: It should have a new response type which mentions `workItems` instead of `products`. we might need to update other endpoints and the docs as well to call `products` as `workItems` and `phase` as `work` when accessed via new `workstream` endpoints.

src/routes/works/update.js
```diff
@@ -0,0 +1,186 @@
...
+          },
+          include: [{ model: models.PhaseProduct, as: 'products' }],
+        })),
+    )
+      .then((allPhases) => {
+        req.log.debug('updated project phase', JSON.stringify(updated, null, 2));
+
+        // emit original and updated project phase information
+        req.app.services.pubsub.publish(
+          EVENT.ROUTING_KEY.PROJECT_PHASE_UPDATED,
```
vikasrohit: We might need to raise a synonym event `PROJECT_WORK_UPDATED` or `WORKSTREAM_WORK_UPDATED` as well for consumers who are listening to events based on new nomenclature of workstreams and works.

src/routes/works/create.js
```diff
@@ -0,0 +1,151 @@
...
+                newProjectPhase.products = [
+                  _.omit(phaseProduct.toJSON(), ['deletedAt', 'deletedBy']),
+                ];
+              });
+          });
+      }))
+      .then(() => {
+        // Send events to buses
+        req.log.debug('Sending event to RabbitMQ bus for project phase %d', newProjectPhase.id);
+        req.app.services.pubsub.publish(EVENT.ROUTING_KEY.PROJECT_PHASE_ADDED,
```
vikasrohit: We might need to raise a synonym event `PROJECT_WORK_ADDED` or `WORKSTREAM_WORK_ADDED` as well for consumers who are listening to events based on new nomenclature of workstreams and works.

maxceem:
@vikasrohit would like to confirm regarding event names for worksteam related stuff. 

We should create synonyms for `EVENT.ROUTING_KEY` keys to send to RabbitMQ or and for `req.app.emit` - that's clear.

What should we do with Bus events? Should we still send `phase` events to Bus when working with workstreams?

vikasrohit:
@maxceem I am okay with keeping the same `phase` event for internal node event, however, I would like to have new events for kafka bus api. We don't need to support RabbitMQ for new events as we are already working on removing RabbitMQ and use the standard kafka bus api instead.

maxceem:
> I would like to hear your suggestions about how create new events while keeping backward compatibility.

I guess the most transparent approach would be to keep sending old events when we are using `phase` related endpoints. And for the `work` related endpoints we can create synonyms of the same BUS events using `work` as a keyword instead of the `phase`. As I udnerstand projects would use either phase or work endpoints so we wouldn't get duplicate synonym events for the same things.

To sum up, I think only the next events need to get `work` synonyms.
```js
 PROJECT_PHASE_TRANSITION_ACTIVE: 'notifications.connect.project.phase.transition.active',
  PROJECT_PHASE_TRANSITION_COMPLETED: 'notifications.connect.project.phase.transition.completed',
  PROJECT_PHASE_UPDATE_PAYMENT: 'notifications.connect.project.phase.update.payment',
  PROJECT_PHASE_UPDATE_PROGRESS: 'notifications.connect.project.phase.update.progress',
  PROJECT_PHASE_UPDATE_SCOPE: 'notifications.connect.project.phase.update.scope',
```
See the full list of events https://github.com/topcoder-platform/tc-project-service/blob/feature%2Fworkstreams/src/constants.js#L91

The rest of the events should stay as they are, no matter if they triggered by `phase` or `work` endpoints.

>  I am okay with keeping the same phase event for internal node event

If we keep internal events then the rest of BUS events would just work. So we can add additional internal events to trigger new `work` BUS events, while keeping old internal events also to support the rest of BUS events which are common for `phases` and `works`.

src/routes/workItems/list.spec.js
```diff
@@ -0,0 +1,225 @@
...
+        .send({ param: body })
+        .expect('Content-Type', /json/)
+        .expect(200)
+        .end((err, res) => {
+          if (err) {
+            done(err);
+          } else {
+            const resJson = res.body.result.content;
+            should.exist(resJson);
+            resJson.should.have.lengthOf(1);
```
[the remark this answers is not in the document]
vikasrohit: @maxceem do you think we can get this fixed in this PR or would you like to create separate PR for fixing this?

swagger.yaml
vikasrohit: @maxceem do you think we can get this fixed in this PR or would you like to create separate PR for fixing this?

maxceem: @vikasrohit I think it's better to fix it before merging in this PR. I'm checking something related to this comment and will update soon.

vikasrohit:
@maxceem Agreed. I think we can keep the two models and their events separate. If `phase` updated (via `phases` endpoints) raise phase events as we are doing now, and if `work` updated (via `works` endpoints) raise work events which are almost same as of `phase` events but with different keywords to reflect the different models

src/routes/workItems/list.spec.js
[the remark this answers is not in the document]
maxceem: @vikasrohit `workItems/list.spec.js` returns a plain list of WorkItems inside `res.body.result.content`. It doesn't have an object with `phases` key also.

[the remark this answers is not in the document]
vikasrohit: Agreed. I was concerned with `/workstreams/{id}/works` endpoint and `/workstreams/{id}/works/{id}/workItems`. Are we validating the keys there?

swagger.yaml
maxceem: @vikasrohit so far we almost never return in response any word like phase. Except for one case.

### Endpoints WITHOUT mentioning if it's work* or phase* related object

- `GET ../phase/{id}`, `GET /work/{id}` - returns one phase/work object, without mentioning if it's phase or work
- `GET ../products/{id}`, `GET /workItems/{id}` - returns one phaseProduct/workItem object, without mentioning if it's phaseProduct or workItem
- `GET ../products`, `GET /workItems` - returns the list of phaseProduct/workItem objects, without mentioning if it's list of phaseProducts or workItems

### Endpoints WITH mention what kind of objects they return

For the list of phases we have TWO endpoints, to retrieve data from ES or DB:
- `GET ../phase`, `GET ../phase/db` - returns the list of phase from ES index or DB. Optionally it can populate phase objects with `products` list if we provide query `?fields=products`

For works we have only one endpoint to get data from DB `GET ../work`. As we agreed in the related issue requirements, we don't use ES index for works (though we add works to ES for backward compatibility for phases endpoints).
- `GET ../work` - returns the list of works from DB. Optionally it can populate phase objects with `products` list if we provide query `?fields=products`. 
So what we have to do here is:
Instead of supporting `products` field, we should support `workItems` field. And we pass it, then this endpoint should populate works with `workItems` lists.

### Swagger file
So far in Swagger, we don't reflect the ability to get `products` inside phases.

Should we add `products` field to the [ProjectPhase](https://github.com/topcoder-platform/tc-project-service/blob/feature/workstreams/swagger.yaml#L4510-L4542) object in swagger?
And create a similar object for Work with `workItem` property?

vikasrohit: > Should we add products field to the ProjectPhase object in swagger?
And create a similar object for Work with workItem property?

I guess, yes. However, we can do that out of this PR.

maxceem: @vikasrohit we read from the DB in all the cases, when we get one work, or when we get multiple works.

So all the next endpoints get results for DB:
GET ../workstreams
GET ../workstreams/:id
GET ../works
GET ../works/:id
GET ../workitems
GET ../workitems/:id

Meanwhile, we still store `works` (phases) and `workitems` (phaseProducts) in the ES to be used in the ONLY next two endpoints:
GET ../phases
GET ../products

maxceem: @vikasrohit regading this:
`GET ../works` - returns the list of works from DB. Optionally it can populate phase objects with products list if we provide query `?fields=products`.

Should we update it return phaseProducts inside `workItems`?
Instead of supporting `products` field, we should support `workItems` field. And we pass it, then this endpoint should populate works with `workItems` list.

vikasrohit: We need to update all READ endpoints to fetch from ES as v5 standards efforts. However, we can do it in a single batch after merging this PR. Also, notice that we need to make sure, that we don't update existing endpoints for this as separate team is working on moving all existing endpoints to v5 standards. So, we just need to make sure that we follow the same standard for our new endpoints introduced by this PR.

Yes, for `GET ../works` endpoint, we need `workItems` as `fields` not `products`. So, that is something we should update before merging this PR.

maxceem:
@vikasrohit I think this PR can now be merged to DEV.

During the final fixes we:
- separated BUS API events for phase and work specific action, while keeping general Project Pan events same
- fixed work-related endpoint to return `worksItems` instead of `products`. (Also fixed phase DB endpoint which was broken).

Things which are left, but can be done on the way:
- Implement some way of initial populating workManagementPermission to make the development easier, as discussed in comments https://github.com/topcoder-platform/tc-project-service/issues/315#issuecomment-513671050
- Put all the new models introduced in this work into ES and update READ endpoint to follow API v5 standard by retrieving data from ES first and fallback to DB

maxceem:
@vikasrohit forgot to mention, the migration script should be run for this PR https://github.com/topcoder-platform/tc-project-service/pull/342/files#diff-1d0d3b3a6b66ebc3d1286381abc57cac

Issue #346, opened 2019-07-17
General permission methods


maxceem:
During working on the Workstreams we had to check permissions based on `allowRule` and `denyRule` as was described in this comment https://github.com/topcoder-platform/tc-project-service/issues/315#issuecomment-503396151. And it was implemented during the challenge in [workManagementForTemplate](https://github.com/topcoder-platform/tc-project-service/pull/342/files#diff-b4c12893f152cab9852018540be29a4d) permission middleware.

As we are going to have more and more places where we would have to restrict permissions based on various set of rules and to avoid issues like this one https://github.com/topcoder-platform/tc-project-service/issues/332 I've moved the logic for checking permissions into general util methods:
- `matchPermissionRule`
- `hasPermission`
- `hasPermissionForProject`

You can check them out in the PR https://github.com/topcoder-platform/tc-project-service/pull/345

The most general method is `hasPermissionForProject: (permission, user, projectId)`:
- `projectId` is project id we should check permission for
- `user` is the user we should check permissions for, usually we would use the current user from the request object `user = req.authUser`
- `permission`

I suggest to be able to define the `permission` in two ways:
- **Full** when permission is defined with `allowRule` and `denyRule`:

    ```js
    {
      allowRule: {
          projectRoles: [],
          topcoderRoles: []
      },
      denyRule: {
          projectRoles: [],
          topcoderRoles: []
      }
    }
    ```
- **Simplified** if we don't need `denyRule` we can deifne directly the inner part of `allowRule` without explicit property `allowRule`:

   ```js
   {
      projectRoles: [],
      topcoderRoles: []
   }
   ```
   This **simplified** permission is equal to a **full** permission:
    ```js
    {
       allowRule: {
         projectRoles: [],
         topcoderRoles: []
       }
    }
    ```

I guess this should cover most of our needs for permissions checking. And after some time we can refactor exitent code to use these methods instead of reimplementing this logic, for example many of [these methods](https://github.com/topcoder-platform/tc-project-service/tree/dev/src/permissions) could be rewritten reusing the general methods.

So far what I see can be done to improve them:
1. Create some way to check that the user is a member of the project with any member role. For now we would have to list all the user roles in `projectRoles` which is too wordy, also when adding a new role we would have to update all permission rules. So far I see two possible ways:
   1. Add new property like `isProjectMember` so the rule would look like:
      ```js
       {
         isProjectMember: true,
         topcoderRoles: []
       }
      ```
[...]

PR #347, opened 2019-07-18
Dev next

maxceem:
@vikasrohit this PR is for your information and confirmation for new changes in the DEV branch:

1. Fixed issue which has been left from CF18 https://github.com/topcoder-platform/tc-project-service/issues/234
2. Implementation of general permission methods with refactored copilotAndAbove permission "middleware" to use new general method for the proof of work. See https://github.com/topcoder-platform/tc-project-service/issues/346 for details.

### DIFF
--- a/src/permissions/constants.js
+++ b/src/permissions/constants.js
@@ -1,0 +1,36 @@
+/**
+ * Definitions of permissions which could be used with util methods
+ * `util.hasPermission` or `util.hasPermissionForProject`.
+ *
+ * We can define permission using two logics:
+ * 1. **WHAT** can be done with such a permission. Such constants may have names like:
+ *    - `VIEW_PROJECT`
+ *    - `EDIT_MILESTONE`
+ *    - `DELETE_WORK`
+ *    and os on.
+ * 2. **WHO** can do actions with such a permission. Such constants **MUST** start from the prefix `ROLES_`, examples:
+ *    - `ROLES_COPILOT_AND_ABOVE`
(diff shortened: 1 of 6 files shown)


vikasrohit:
Thanks @maxceem. For the permission logic, do you think the util you have written now, is potentially candidate to be moved to the [core library](https://github.com/appirio-tech/tc-core-library-js/blob/master/lib/authorizer/index.js) itself so that others can take benefit of the same logic and we have streamlined way to define permissions?

maxceem:
> Thanks @maxceem. For the permission logic, do you think the util you have written now, is potentially candidate to be moved to the core library itself so that others can take benefit of the same logic and we have streamlined way to define permissions?

2 of 3 methods don't depend on the DB, so they for sure can be moved to the core library (potentially). Also, I wrote them by keeping in mind that we are trying to implement simmilar methods in Connect App https://github.com/appirio-tech/connect-app/issues/2841.

Anyway, the most important thing is not methods itself, but the format to define permissions. It would be great if we could use the same permission format across all the repos.

So I guess the plan here could be like:
1. use new general methods and unified permission format in the Project Service
2. update general methods to use the same unified permission format in Connect App
3. see how it works, if there any limitations, if it's comfortable to use them, if they are pron to any errors and so on.
4. adjust methods and formats if neccessary
5. move the common permission methods to the core library

vikasrohit:
Sounds like a plan. Go ahead.

PR #348, opened 2019-07-18
implement scope change request workflow

vigneshTheDev:
Implements the task in the issue: https://github.com/appirio-tech/connect-app/issues/2970

### DIFF
--- a/src/routes/projects/update.js
+++ b/src/routes/projects/update.js
@@ -82,11 +82,31 @@ const updateProjectValdiations = {
   },
 };
 
+/**
+ * Gets scopechange fields either from
+ * "template.scope" (for old templates) or from "form.scope" (for new templates).
+ *
+ * @param {Object} project The project object
+ *
+ * @returns {Array} - the scopeChangeFields
+ */
+const getScopeChangeFields = (project) => {
+  const scopeChangeFields = _.get(project, 'template.scope.scopeChangeFields');
+
+  const getFromForm = (_project) => {
+    const formRef = _.get(_project, 'template.form');
+    const forms = _.get(_project, 'forms');
+    const form = formRef && _.find(forms, formRef);
+
+    return _.get(form, 'scope.scopeChangeFields');
+  };
+
+  return scopeChangeFields || getFromForm(project);
+};
+
 const isScopeUpdated = (existingProject, updatedProps) => {
-  const template = existingProject.template;
-  if (!template) return false;
-  const scopeFields = template.scope.scopeChangeFields;
-  // TODO handle new refactored scope form
+  const scopeFields = getScopeChangeFields(existingProject);
+
   if (scopeFields) {
     for (let idx = 0; idx < scopeFields.length; idx += 1) {
       const field = scopeFields[idx];
--- a/src/routes/projects/update.js
+++ b/src/routes/projects/update.js
@@ -92,12 +92,7 @@ const updateProjectValdiations = {
  */
 const getScopeChangeFields = (project) => {
   const scopeChangeFields = _.get(project, 'template.scope.scopeChangeFields');
-
-  const getFromForm = (_project) => {
-    const form = _project.form;
-
-    return _.get(form, 'config.scopeChangeFields');
-  };
+  const getFromForm = _project => _.get(_project, 'template.form.config.scopeChangeFields');
 
   return scopeChangeFields || getFromForm(project);
 };
@@ -203,19 +198,13 @@ module.exports = [
           return Promise.reject(err);
         }
         if (!_prj.templateId) return Promise.resolve({ _prj });
-        return models.ProjectTemplate.findById(_prj.templateId, { raw: true })
+        return models.ProjectTemplate.getTemplate(_prj.templateId)
         .then(template => Promise.resolve({ _prj, template }));
       })
       .then(({ _prj, template }) => {
-        const formRef = template.form;
-        return formRef ? models.Form.findOne({ where: formRef })
-          .then(form => ({ _prj, template, form })) : { _prj, template };
-      })
-      .then(({ _prj, template, form }) => {
         project = _prj;
         previousValue = _.clone(project.get({ plain: true }));
         previousValue.template = template;
-        previousValue.form = form;
         // run additional validations
         const validationErrors = validateUpdates(previousValue, updatedProps, req);
         if (validationErrors.length > 0) {
--- a/src/routes/projects/update.js
+++ b/src/routes/projects/update.js
@@ -94,11 +94,9 @@ const getScopeChangeFields = (project) => {
   const scopeChangeFields = _.get(project, 'template.scope.scopeChangeFields');
 
   const getFromForm = (_project) => {
-    const formRef = _.get(_project, 'template.form');
-    const forms = _.get(_project, 'forms');
-    const form = formRef && _.find(forms, formRef);
+    const form = _project.form;
 
-    return _.get(form, 'scope.scopeChangeFields');
+    return _.get(form, 'config.scopeChangeFields');
   };
 
   return scopeChangeFields || getFromForm(project);
@@ -209,9 +207,15 @@ module.exports = [
         .then(template => Promise.resolve({ _prj, template }));
       })
       .then(({ _prj, template }) => {
+        const formRef = template.form;
+        return formRef ? models.Form.findOne({ where: formRef })
+          .then(form => ({ _prj, template, form })) : { _prj, template };
+      })
+      .then(({ _prj, template, form }) => {
         project = _prj;
         previousValue = _.clone(project.get({ plain: true }));
         previousValue.template = template;
+        previousValue.form = form;
         // run additional validations
         const validationErrors = validateUpdates(previousValue, updatedProps, req);
         if (validationErrors.length > 0) {
--- a/src/routes/scopeChangeRequests/create.js
+++ b/src/routes/scopeChangeRequests/create.js
@@ -55,11 +55,11 @@ module.exports = [
 
       // If the project is not frozen yet, the changes can be saved directly into projects db.
       // Scope change request workflow is not required.
-      const statesesForNonFrozenProjects = [PROJECT_STATUS.DRAFT, PROJECT_STATUS.IN_REVIEW];
-      if (statesesForNonFrozenProjects.indexOf(project.status) > -1) {
+      const statusesForNonFrozenProjects = [PROJECT_STATUS.DRAFT, PROJECT_STATUS.IN_REVIEW];
+      if (statusesForNonFrozenProjects.indexOf(project.status) > -1) {
         const err = new Error(
           `Cannot create a scope change request for projects with statuses: ${
-            statesesForNonFrozenProjects.join(', ')}`);
+            statusesForNonFrozenProjects.join(', ')}`);
         err.status = 403;
         return Promise.reject(err);
       }
--- a/src/routes/scopeChangeRequests/create.js
+++ b/src/routes/scopeChangeRequests/create.js
@@ -3,7 +3,7 @@ import Joi from 'joi';
 import validate from 'express-validation';
 import { middleware as tcMiddleware } from 'tc-core-library-js';
 import util from '../../util';
-import { SCOPE_CHANGE_REQ_STATUS, PROJECT_MEMBER_ROLE } from '../../constants';
+import { SCOPE_CHANGE_REQ_STATUS, PROJECT_MEMBER_ROLE, PROJECT_STATUS } from '../../constants';
 import models from '../../models';
 
 /**
@@ -15,8 +15,6 @@ const createScopeChangeRequestValidations = {
   body: {
     param: Joi.object()
       .keys({
-        title: Joi.string().max(90),
-        description: Joi.string().max(255),
         oldScope: Joi.object(),
         newScope: Joi.object(),
       }),
@@ -29,8 +27,6 @@ module.exports = [
   permissions('project.edit'),
   (req, res, next) => {
     const projectId = _.parseInt(req.params.projectId);
-    const title = _.get(req, 'body.param.title');
-    const description = _.get(req, 'body.param.description');
     const oldScope = _.get(req, 'body.param.oldScope');
     const newScope = _.get(req, 'body.param.newScope');
     const members = req.context.currentProjectMembers;
@@ -38,8 +34,6 @@ module.exports = [
         m => m.userId === req.authUser.userId && m.role === PROJECT_MEMBER_ROLE.CUSTOMER));
 
     const scopeChange = {
-      title,
-      description,
       oldScope,
       newScope,
       status: isCustomer ? SCOPE_CHANGE_REQ_STATUS.APPROVED : SCOPE_CHANGE_REQ_STATUS.PENDING,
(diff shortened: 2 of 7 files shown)


src/routes/scopeChangeRequests/update.js
```diff
@@ -72,9 +68,18 @@ module.exports = [
         err.status = 401;
         return next(err);
       }
-      return models.ScopeChangeRequest.update(updatedProps, {
-        where: { id: requestId },
-      }).then((_updatedReq) => {
+
+      return (
+        updatedProps.status === SCOPE_CHANGE_REQ_STATUS.ACTIVATED
+          ? models.Project.update({ details: scopeChangeReq.newScope }, { where: { id: projectId } })
```
vikasrohit: As I said in review of front end PR, we need to consider the case where some fields(which does not count for scope change) of the `details` json are updated after the scope change request is created and before it is activated.

src/routes/projects/update.js
```diff
@@ -82,11 +82,31 @@ const updateProjectValdiations = {
...
+ * @param {Object} project The project object
+ *
+ * @returns {Array} - the scopeChangeFields
+ */
+const getScopeChangeFields = (project) => {
+  const scopeChangeFields = _.get(project, 'template.scope.scopeChangeFields');
+
+  const getFromForm = (_project) => {
+    const formRef = _.get(_project, 'template.form');
+    const forms = _.get(_project, 'forms');
```
vikasrohit: where we are setting `forms` to the `_project`? I mean I don't see we are fetching the forms anywhere in our code when fetching the `template`.

src/routes/scopeChangeRequests/update.js
vikasrohit: And it would be great if we can add unit tests for this scenario as well.

vigneshTheDev: Hi @vikasrohit, I have a query here. In the scope change request flow, we're not filtering/extracting the fields mentioned in scopeChangeFields from details json. But based on this comment, we have to filter right? If there are changes in the fields that are not part of scopeChangeFields, should we call update Project endpoint to directly save these fields from the UI?

vikasrohit: We are doing that filtering here https://github.com/topcoder-platform/tc-project-service/blob/82994ca9f0b17b15dbc4b91a184647a5b00280df/src/routes/projects/update.js#L107. The method returns true only when there exists at least one field from the `scopeChangeFields` list which is updated. If all the fields that are updated are not listed in the `scopeChangeFields`, then are allowing the project to be updated no matter what is the status of the project (except `completed`).
This brings another point for the front end stuff. I think we are not checking `scopeChangeFields` on the front end to do the correct operation based on the fields updated. I think on the front end we always assuming that it would create a scope change request when project is in `reviewed` or above status, even if there is no field from the `scopeChangeFields` is updated by the user. We can live with that for the purpose of this task, however, it would be great if we can handle that after merging these PRs. Let me know, what do you think about that.

src/routes/scopeChangeRequests/create.js
```diff
@@ -48,12 +42,48 @@ module.exports = [
...
+    .then((project) => {
+      if (!project) {
+        const err = new Error(`Project with id ${projectId} not found`);
+        err.status = 404;
+        return Promise.reject(err);
+      }
+
+      // If the project is not frozen yet, the changes can be saved directly into projects db.
+      // Scope change request workflow is not required.
+      const statesesForNonFrozenProjects = [PROJECT_STATUS.DRAFT, PROJECT_STATUS.IN_REVIEW];
```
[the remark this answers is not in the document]
vigneshTheDev: This is done

src/routes/projects/update.js
vigneshTheDev: I've fixed this. I'll test it a couple of rounds and push by EOD.

src/routes/scopeChangeRequests/update.js
vigneshTheDev: Fixed this. Also added the unit test case. 

> I think on the front end we always assuming that it would create a scope change request when project is in reviewed or above status, even if there is no field from the scopeChangeFields is updated by the user.

This, we can take up after this PR

src/routes/projects/update.js
vigneshTheDev: I pushed the changes for this.

src/routes/scopeChangeRequests/update.js
vikasrohit: Thanks.

> There are fields in the sections that are not coming under details json altogether. We have to handle those also I guess. 

IMO, if there is any field, that is mentioned in `scopeChangeFields` list, is updated by user, we should deny the request with clear error that `<field_1>, <field_2> can not be updated because they change the scope of the project. Please create scope change request to update them.`.  User has to remove such fields from the request to update any other field, no matter if they are in `details` json or not. I think this is exactly(except the text of the error message) what we are doing right now, correct?

src/routes/projects/update.js
```diff
@@ -209,9 +207,15 @@ module.exports = [
         .then(template => Promise.resolve({ _prj, template }));
       })
       .then(({ _prj, template }) => {
+        const formRef = template.form;
+        return formRef ? models.Form.findOne({ where: formRef })
```
vikasrohit: @vigneshTheDev can we create and use a class method (we may call it `getTemplate`) of the model `ProjectTemplate`, which fetches the `Form` object as well using [joins with json fields](https://dba.stackexchange.com/questions/83932/postgresql-joining-using-jsonb)? That would be need any how in future to make the refactoring of templates into forms, transparent.

src/routes/scopeChangeRequests/update.js
```diff
@@ -71,7 +91,7 @@ module.exports = [
 
       return (
         updatedProps.status === SCOPE_CHANGE_REQ_STATUS.ACTIVATED
-          ? models.Project.update({ details: scopeChangeReq.newScope }, { where: { id: projectId } })
+          ? updateProjectDetails(scopeChangeReq.newScope, projectId)
           : Promise.resolve()
       )
       .then(() => scopeChangeReq.update(updatedProps))
```
vikasrohit: @vigneshTheDev Just noticed that we are not updating the project in ES index, when scope change request is activated.

vikasrohit: @vigneshTheDev if that seems to increase the scope of our agreed requirements, we can take care of it in a separate task, as practically we don't read `project.details` from ES yet. However, with our efforts towards v5 standards, we have to read it from ES in very near future.

src/routes/scopeChangeRequests/update.js
vigneshTheDev: Yes Correct. In the backend we're doing it. I was talking about the UI perspective. The UI always raises scope change request for projects with status reviewed and later. Suppose, I change the title. Which is not coming under details.json. But we're not calling update project endpoint if the project is in reviewed or later status.

src/routes/scopeChangeRequests/update.js
vigneshTheDev: Ok. Let me check and get back to you on this. If it's simple, I'll add it. Otherwise, we can take it as a separate task.

src/routes/scopeChangeRequests/update.js
vikasrohit: Okay. I think we are good for now as long as user can remove the fields which changes the scope from their update request and able to update the project e.g. a simple patch request with project title should work just fine.
Yep, definitely we need to update our UI for letting user do such updates and I think some where on this PR we discussed that we have to do that as separate task.

vikasrohit: Created [issue](https://github.com/topcoder-platform/tc-project-service/issues/351) for tracking those efforts on front end. Please feel free to add more details into it.

src/routes/projects/update.js
vigneshTheDev: Joining merges the form and template objects right? like this: https://i.imgur.com/HyXeFeE.png
Shall I extract the existing logic into the class method `getTemplate` instead?

vikasrohit: Yes for both questions. Just a performance thing, as per the answer in the link I shared in my previous comment for joining with JSON fields, may be we should use `->` instead of `->>`.

> Use the -> operator instead of ->> in the reference to children. The way you have it, you'd first cast json / jsonb to text and then back to json.

vigneshTheDev: If I understood it correctly, in the answer you've shared, they're dealing with an array value. i.e., p.data->'children' is an array. He says, it's not efficient to convert it to a string and back to a json array. But in our case, we're dealing with text fields. i.e., template.form->key and template.form->version. I think in our case it makes sense to use ->>. Please correct me if I'm wrong. 

Actually, my original question was this:
What should we return from getTemplate method?
Do we want to return the template like this:
```js
{
  id: 1,
  name: '',
  form: {
    config: {
      id: 1,
    }
  }
}
```

Or like this:
```js
{
  id: 1,
  name: '',
  config: {
    id: 1,
  }
}
```

Note that in the 2nd version, form config is directly put inside the template object. I guess joining would result like this.

If we want in the first way, I was suggesting the current implementation as it is except it'll be extracted as a class method (without using joins).

Also, if a form with same version has multiple revisions, will we have revision number in template.form?

src/routes/scopeChangeRequests/update.js
vigneshTheDev: @vikasrohit, this is done.

src/routes/projects/update.js
vigneshTheDev: Hi @vikasrohit, any thoughts on this one?

vikasrohit: I like the first one where we have `form` as key of the `template` object. Lets go with your current implementation but in a class method so that we can later on refactor it.
> Also, if a form with same version has multiple revisions, will we have revision number in template.form?
We should pick the latest revision.

vikasrohit: For `->` vs `->>` lets leave it for now as you have it as of now.

src/routes/scopeChangeRequests/update.js
```diff
@@ -32,16 +33,31 @@ const updateScopeChangeRequestValidations = {
...
       const err = new Error('Project not found');
       err.status = 404;
       return Promise.reject(err);
     }
 
     const updatedDetails = _.merge({}, project.details, newScope);
-    return project.update({ details: updatedDetails });
+    return project.update({ details: updatedDetails }).then((updatedProject) => {
+      const updated = updatedProject.get({ plain: true });
+      const original = _.omit(previousValue, ['deletedAt']);
```
vikasrohit: I think we usually exclude the `deletedBy` as well from being indexing in the ES.

src/routes/projects/update.js
vigneshTheDev: This is done

src/routes/scopeChangeRequests/update.js
vigneshTheDev: This is done.

vikasrohit:
Thanks @vigneshTheDev. Just to confirm, we don't need to run any SQL migration script for this PR, right?

vigneshTheDev:
@vikasrohit No we don't need to run any scripts

vikasrohit:
@vigneshTheDev some tests are failing. Do you think it is only env issue?

vigneshTheDev:
@vikasrohit, I found that 15 tests are failing in  _LIST Project_ (src\routes\projects\list.spec.js). I'm looking into that. Can you please share which tests/how many tests are failing for you?

vigneshTheDev:
@vikasrohit, I ran npm run sync:db and npm run sync:es after setting NODE_ENV=test. All the tests are passing after that.

vikasrohit:
@vigneshTheDev I am still getting the [errors](https://circleci.com/gh/topcoder-platform/tc-project-service/1708) in unit tests. Could you please try rebuilding your local docker so that you have a new database in your local setup and then run the unit tests?

vikasrohit:
It worked @vigneshTheDev now. Seems like caching issue with Circle CI.

vikasrohit:
@vigneshTheDev Could you please add migration script for adding the scope_change_requests table?

vigneshTheDev:
@vikasrohit, Added the migration script. Please find the PR here: https://github.com/topcoder-platform/tc-project-service/pull/355

vikasrohit:
Thanks @vigneshTheDev 

I deployed the changes on the dev. I am able to submit a change request, however, I am not able to cancel it because it is throwing some runtime error and killing the node process.

```
start request { method: 'PATCH',\n  url: '/v4/projects/8245/scopeChangeRequests/4' }
finding scope change 4
Stacktrace (bbbbbbbb-bbbbbbbb) 0x4c818302271 (nil):
==== JS stack trace =========================================
Security context: 0x11317961bbd9 <JS Object>#0#
2: /* anonymous */ [/usr/src/app/node_modules/continuation-local-storage/context.js:88] [bytecode=0x2ffa5c483d09 offset=93](this=0x337435de5311 <a Model with map 0x21d9a3b6fb79>#1#)
3: arguments adaptor frame: 1->0
5: tryCatcher [/usr/src/app/node_modules/bluebird/js/release/util.js:16] [bytecode=0x386d67891581 offset=30](this=0x337435de5311 <a Model with map 0x21d9a3b6fb79>#1#)
6: arguments adaptor frame: 1->0
7: _settlePromiseFromHandler [/usr/src/app/node_modules/bluebird/js/release/promise.js:~496] [pc=0xe0f94227c05](this=0x9e31d754939 <a Promise with map 0xa444281ab41>#2#,handler=0x9e31d754891 <JS Function (SharedFunctionInfo 0x17e123f45799)>#3#,receiver=0x337435de5311 <a Model with map 0x21d9a3b6fb79>#1#,value=0x9e31d754741 <JS Object>#4#,promise=0x9e31d7548f9 <a Promise with map 0xa444281ab41>#5#)
8: _settlePromise [/usr/src/app/node_modules/bluebird/js/release/promise.js:~542] [pc=0xe0f94247316](this=0x9e31d754939 <a Promise with map 0xa444281ab41>#2#,promise=0x9e31d7548f9 <a Promise with map 0xa444281ab41>#5#,handler=0x9e31d754891 <JS Function (SharedFunctionInfo 0x17e123f45799)>#3#,receiver=0x337435de5311 <a Model with map 0x21d9a3b6fb79>#1#,value=0x9e31d754741 <JS Object>#4#)
9: _callback [/usr/src/app/node_modules/bluebird/js/release/async.js:~16] [pc=0xe0f944b9fc7](this=0x9e31d754979 <an Immediate with map 0x37b86187f821>#6#)
10: runCallback(aka runCallback) [timers.js:781] [pc=0xe0f9459bc4a](this=0x113179602241 <undefined>,timer=0x9e31d754979 <an Immediate with map 0x37b86187f821>#6#)
12: tryOnImmediate(aka tryOnImmediate) [timers.js:743] [bytecode=0x386d678fd131 offset=82](this=0x113179602241 <undefined>,immediate=0x9e31d754979 <an Immediate with map 0x37b86187f821>#6#,oldTail=0x9e31d754979 <an Immediate with map 0x37b86187f821>#6#)
13: _immediateCallback(aka processImmediate) [timers.js:714] [pc=0xe0f9459b725](this=0x244ef6c87859 <a process with map 0x37b86180de61>#7#)
==== Details ================================================
[2]: /* anonymous */ [/usr/src/app/node_modules/continuation-local-storage/context.js:88] [bytecode=0x2ffa5c483d09 offset=93](this=0x337435de5311 <a Model with map 0x21d9a3b6fb79>#1#) {
```

Do you think you know the reason for this?

vikasrohit:
Same thing happens with `Reject` as well.

vigneshTheDev:
@vikasrohit, It's working fine in my system. Let me check once again.

vikasrohit:
It might be taking unusual memory which server does not have capacity to fulfill but your local machine has enough of that. We are running the service as ECS task which has very limited memory and CPU e.g. dev env has about 1024 vCPU and 2048 MB of memory.

vikasrohit:
@vigneshTheDev I just removed the logging of the found scope change request object here https://github.com/topcoder-platform/tc-project-service/blob/feature/scope_change_workflow/src/routes/scopeChangeRequests/update.js#L83 which seems to fix the issue. It might be causing one of the resources to exhaust because of recursion while logging full sequelize object.

vigneshTheDev:
@vikasrohit got it... good that you caught it. Thanks for the update!

PR #350, opened 2019-07-23
66-FF-1 Workstreams - Project Service

tc-mfikria:


maxceem:
@vikasrohit in this PR we are implementing support of sending separate BUS API events when we manage `phases` and `works`. I like the approach of reusing the same code so in case we do any changes later for one of the endpoints types we wouldn't miss updating the same for anther. Just would like you to have quick a look at the general approach here. If you see any disadvantages.

vikasrohit:
@maxceem I don't see any problem with the approach.

PR #354, opened 2019-07-29
Feature/scope change workflow

vikasrohit:
Scope Change Workflow

fyi @maxceem @RishiRajSahu


vikasrohit:
Merging even when test cases are failing to check if it is branch specific issue. Build worked fine couple of time before starting to fail again.

PR #355, opened 2019-07-29
Add migration scripts

vigneshTheDev:


migrations/20190729_scope_change_requests.sql
```diff
@@ -0,0 +1,40 @@
...
+--   scope_change_requests
+--
+
+CREATE TABLE scope_change_requests
+(
+    id bigint NOT NULL,
+    "projectId" bigint NOT NULL,
+    "oldScope" json NOT NULL,
+    "newScope" json NOT NULL,
+    status character varying(45) COLLATE pg_catalog."default" NOT NULL,
```
vikasrohit: What does the `COLLATE pg_catalog."default"` does here?

migrations/20190729_scope_change_requests.sql
```diff
@@ -0,0 +1,40 @@
...
+    "deletedAt" timestamp with time zone,
+    "createdAt" timestamp with time zone,
+    "updatedAt" timestamp with time zone,
+    "approvedAt" timestamp with time zone,
+    "deletedBy" integer,
+    "createdBy" integer NOT NULL,
+    "updatedBy" integer NOT NULL,
+    "approvedBy" integer,
+    CONSTRAINT scope_change_requests_pkey PRIMARY KEY (id),
+    CONSTRAINT "scope_change_requests_projectId_fkey" FOREIGN KEY ("projectId")
```
vikasrohit: @vigneshTheDev we usually add the primary and foreign key constraints via separate DDLs instead of wrapping in the main CREATE script, could you please refer any one of the existing ones and adapt it for the same?

vigneshTheDev: @vikasrohit This is done.

migrations/20190729_scope_change_requests.sql
vigneshTheDev: Looks like I left some noise after generating the script from pg admin. Fixed it.

Issue #357, opened 2019-07-31
[Workstreams] create topics for works


maxceem:
This is part of https://github.com/topcoder-platform/tc-project-service/issues/315

Currently, for works we create only one topic with tag like "phase#1231". Instead for works, we should create two topics, with tags like "work#1234-details" and "work#1234-requirements".

vikasrohit:
@maxceem I am thinking of doing this by consuming the bus event of project creation. And as per new v5 standards, we are creating separate processors for any work that is done by consuming events from bus e.g. https://github.com/topcoder-platform/project-processor-es. Do you think we can write yet another processor for any event based operations which we need to perform within project service. I mean we already have processor for ES indexing of project data and another for legacy operations like creating direct project, so we can create another processor for our future event based work which we are doing right now within project service or project bus consumers. May be we can call this new processor as project-processor-plan which processes events related to the project plan and in this case, it would create a topic on work creation event. We can use this same processor for creating the default work streams for a new project.

maxceem:
> Do you think we can write yet another processor for any event based operations which we need to perform within project service.

1.  I think we can do it. The only concern I have is, do you know how fast Bus events are processed in our system? I wonder if it's possible that when we create a new project and go to the project Dashboard, the event is not yet processed, and workstreams with topics are not yet created. This would force the user to reload the page several times and wait until the data is updated.
    We have such a situation with the challenges in community app. As far as I know it's connected with some legacy systems though it's super frustrating. So I just want to make sure that we don't have such situations for Connect App for the new code we are creating.

2. Another thought I have regarding this thing. Is how we can enforce data consistency if something went wrong during some processing. Though for workstreams/posts creating that doesn't sound supercritical. But is there any good practice we are using for such cases? Or we just don't put the critical actions in processors?

3. Meanwhile, for the sake of the fast release, as I'm hoping to release it with comments. Do you think it would be ok, to make this task first using the old approach with Rabbit events?
    And later migrate to the processor. Also, note, that I've already done workstreams creating using just synced code inside project creating endpoint, same like for phases. So we can after migrate several features to the processor.

vikasrohit:
[...]

PR #369, opened 2019-08-16
Feature/topics for works

maxceem:
Implementation for https://github.com/topcoder-platform/tc-project-service/issues/357

Creating 2 topics for works to show comments on Details and Requirements tabs.

Notes:
- Keep the same code base between phases and works, so later we can update the common functionality in the same place
- Implemented unit tests, to verify that for phases we create/update/delete one topic as before. And that for new work endpoints we create/update/delete two topics.

Tested it locally and also have unit tests. So I think it's safe to merge and continue testing on DEV.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,6 +48,7 @@ jobs:
         - POSTGRES_USER: circle_test
         - POSTGRES_DB: circle_test
       - image: elasticsearch:2.3
+      - image: rabbitmq:3-management
     environment:
       DB_MASTER_URL: postgres://circle_test:@127.0.0.1:5432/circle_test
       AUTH_SECRET: secret
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,7 @@
     "mocha": true
   },
   "rules": {
-    "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.js", "**/*.spec.js", "**/serviceMocks.js"]}],
+    "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.js", "**/*.spec.js", "src/tests/*.js"]}],
     "max-len": ["error", { "ignoreComments": true, "code": 120 }],
     "valid-jsdoc": ["error", {
       "requireReturn": true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -3308,6 +3308,11 @@
         "object-assign": "^4.0.1"
       }
     },
+    "file-uri-to-path": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/file-uri-to-path/-/file-uri-to-path-1.0.0.tgz",
+      "integrity": "sha512-0Zt+s3L7Vf1biwWZ29aARiVYLx7iMGnEUl9x33fbB/j3jR81u/O2LbqK+Bm1CDSNDKVtJ/YjwY7TUd5SkeLQLw=="
+    },
     "filename-regex": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/filename-regex/-/filename-regex-2.0.1.tgz",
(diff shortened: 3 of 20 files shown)


vikasrohit:
@maxceem It seems something is changed in terms of unit tests setup, 138 unit tests are failing on CircleCI.

maxceem:
@vikasrohit yep, checking...

maxceem:
@vikasrohit Unit tests for this PR are fixed now. The main issue was that `CircleCI` config didn't have RabbitMQ set up. So I've updated `.circleci/config.yml`.

vikasrohit:
Thanks @maxceem However, I am little confused how the build was working fine earlier then?

maxceem:
The difference between before and now is that before we only mock RabbitMQ. But for new tests regarding creating topics for phases and works, we had to un-mock RabbitMQ for these tests, i. e. we have to use real RabbitMQ setup, as topics are created by listening to RabbitMQ events.

vikasrohit:
Thanks for the info @maxceem Now when we have the rabbitmq working in the CI/CD, we can add unit tests for more event based processing in this service. Could you please log an issue for that, we would take care of that later.

maxceem:
Issue created https://github.com/topcoder-platform/tc-project-service/issues/371

vikasrohit:
@maxceem one test is failing after merging it to dev.

```
 1) projectUpdatedKafkaHandler integration "before each" hook for "should throw exception when project not found by id":
     [mapper_parsing_exception] failed to parse [doc.lastActivityAt]
  :: {"path":"/projects_test/projectV4/1","query":{},"body":"{\"doc\":{\"createdAt\":\"2019-08-19T10:14:13.000Z\",\"updatedAt\":\"2019-08-19T10:14:13.000Z\",\"terms\":[],\"version\":\"v3\",\"id\":1,\"type\":\"generic\",\"billingAccountId\":1,\"name\":\"test1\",\"description\":\"test project1\",\"status\":\"draft\",\"details\":{},\"createdBy\":1,\"updatedBy\":1,\"lastActivityAt\":\"1970-01-01T00:00:00.001Z\",\"lastActivityUserId\":\"1\",\"directProjectId\":null,\"external\":null,\"bookmarks\":null,\"utm\":null,\"estimatedPrice\":null,\"actualPrice\":null,\"challengeEligibility\":null,\"cancelReason\":null,\"templateId\":null,\"deletedAt\":null,\"deletedBy\":null}}","statusCode":400,"response":"{\"error\":{\"root_cause\":[{\"type\":\"mapper_parsing_exception\",\"reason\":\"failed to parse [doc.lastActivityAt]\"}],\"type\":\"mapper_parsing_exception\",\"reason\":\"failed to parse [doc.lastActivityAt]\",\"caused_by\":{\"type\":\"number_format_exception\",\"reason\":\"For input string: \\\"1970-01-01T00:00:00.001Z\\\"\"}},\"status\":400}"}
      at respond (node_modules/elasticsearch/src/lib/transport.js:289:15)
      at checkRespForFailure (node_modules/elasticsearch/src/lib/transport.js:248:7)
      at HttpConnector.<anonymous> (node_modules/elasticsearch/src/lib/connectors/http.js:164:7)
      at IncomingMessage.wrapper (node_modules/elasticsearch/node_modules/lodash/index.js:3095:19)
      at endReadableNT (_stream_readable.js:1055:12)
      at node_modules/async-listener/glue.js:188:31
      at _combinedTickCallback (internal/process/next_tick.js:138:11)
      at process._tickCallback (internal/process/next_tick.js:180:9)
```

maxceem:
@vikasrohit works on the third attempt:

image

vikasrohit:
It failed again. :(

maxceem:
The last fail was strange as it looks like there is RabbitMQ in the docker, it shouldn't happen as I've already configured the CircleCI for RabbitMQ. I re-run again and it works.

vikasrohit:
May be we need to add some time delay to wait for RabbitMQ to start properly before actual tests start.

vikasrohit:
I tried to rerun it and it failed again. So, I think there is something definitely wrong here. And the error is consistent:
`failed to parse [doc.lastActivityAt]\",\"caused_by\":{\"type\":\"number_format_exception\",\"reason\":\"For input string: \\\"1970-01-01T00:00:00.001Z\\\"\"`

maxceem:
Yeah, this error is not because of CircleCI setup. Cannot reproduce it locally even once, but will try to search for possible reasons.

vikasrohit:
May be this error happens when we rebuild the docker images (used in local setup) because on circle ci every time it fetches the docker images so essentially it starts from scratch on CircleCI while on our local setup, we have the basic setup already working for every test case run. It caused issues in past as well when we were asserting the ids of the new records in unit tests the reason was CircleCI has primary key sequences reset on every run while our local setup does not.

maxceem:
Hmm, so far there two types of failing build.

1. The error which happened before [with multiple tests failed](https://circleci.com/gh/topcoder-platform/tc-project-service/1816) could be caused by CircleCI, as actually it shouldn't happen after I've added RabbitMQ to circle config.

2. The error from the last failed build mentioned in this [comment](https://github.com/topcoder-platform/tc-project-service/pull/369#issuecomment-522524075) most likely is caused by tests interfering with each other. Will check it out. 

If the first case with multiple errors would happen again, would have to check it out also.

vikasrohit:
1. the link to the build seems wrong @maxceem as there is only 1 test failing in that build. I think you meant [this](https://circleci.com/gh/topcoder-platform/tc-project-service/1802) one. If you are concerned about this one,  I think we are good with that as now we have the rabbitmq in circle ci.
2. Yep, this is one we have to take care of.

PR #387, opened 2019-10-01
New user roles

maxceem:
Project Service support of new roles as per https://github.com/appirio-tech/connect-app/issues/3298


maxceem:
> Just to confirm @maxceem, when a user with new roles in trying to join a project using Join button, they would join as standard Manager role, not with their specific role, right? (Ideally we want them to join project with their specific user role only)

Right. So far like that. 
- If a user with a new user role like “Business Development Representative”, “Presales”, “Account Executive” click Join button, he will join as “accound_manager”
- If a user with a new manager roles like “Program Manager”, “Solution Architect”, “Project Manager” clicks "Join" button, he will join as "manager" by default. Though such a user will see a dialog where he can choose more specific role:

   image

> Also, when some else is inviting a user with new roles or editing roles of such users, they would be able to assign these new roles to their specific roles along with the regular Manager role, right? (Ideally we want them only to be able to assume the user role they have)

Right. So far we give users a freedom to assign any role of the same level. Like “Solution Architect” can become a "Manager" inside project.

vikasrohit:
One more observation @maxceem when a `Connect Manager` is trying to join the project, as Program Manager, using the `Join` button, it is erroring out saying that user needs `Program Manager` role to join project in the same capacity. Just wanted to know, if you are also on the same page for this.

maxceem:
> One more observation @maxceem when a Connect Manager is trying to join the project, as Program Manager, using the Join button, it is erroring out saying that user needs Program Manager role to join project in the same capacity. Just wanted to know, if you are also on the same page for this.

Thanks for pointing that out. I've noticed, that there is some incosistency in how we let users invite any user with any role, while when users join the project themselves, the rules are more strict.

I've create a new issue to track what improvements we can do for the new user roles https://github.com/appirio-tech/connect-app/issues/3345. Most of them can be done client-side only. While server-side issues could be releases later, as they are not blocking, server-side changes mostly only need to restrict user actions on the backend.

PR #407, opened 2019-11-07
[PROD] Hotfix /milestone completion date

maxceem:
Supporting fixes for Connect App ticket https://github.com/appirio-tech/connect-app/issues/3210

- only admins should be able to edit `completionDate` and `actualStartDate`
- fix issue: it was validated against `startDate` instead of `actualStartDate` so `comletionDate` couldn’t be earlier than `startDate` which is not important when there is `actualStartDate` which could be earlier
- there was another error even if `comletionDate` passes the validation: `comletionDate.diff is not a function`


maxceem:
@vikasrohit I've created this HOTFIX for PROD by cherry-picking some commits from DEV. 
It works good on DEV.

PR #413, opened 2019-11-22
Project Members Traits

yoution:
[challenge](https://www.topcoder.com/challenges/30107353)


maxceem:
Thanks for the PR.

I've already pushed the results of the challenge from the git patch, as needed it urgently, so I'm closing this PR as the changes are already there.

PR #414, opened 2019-11-28
feat: project member & invites endpoint with additional fields

dpkshrma:
Added/updated following endpoints to get user information with additional data:
1. project member by id
2. list of all invites inside the project
3. authenticated user invite details

### DIFF
--- a/src/routes/projectMemberInvites/list.js
+++ b/src/routes/projectMemberInvites/list.js
@@ -27,10 +27,7 @@ module.exports = [
         fields = req.query.fields.split(',');
       }
       const projectId = _.parseInt(req.params.projectId);
-      const invites = await models.ProjectMemberInvite.findAll({
-        where: { projectId },
-        raw: true,
-      });
+      const invites = await models.ProjectMemberInvite.getPendingInvitesForProject(projectId);
       const invitesWithDetails = await util.getObjectsWithMemberDetails(invites, fields, req);
       return res.json(util.wrapResponse(req.id, invitesWithDetails));
     } catch (err) {
--- a/src/routes/projectMemberInvites/list.js
+++ b/src/routes/projectMemberInvites/list.js
@@ -26,18 +26,12 @@ module.exports = [
       if (req.query.fields) {
         fields = req.query.fields.split(',');
       }
-      const memberFields = _.keys(models.ProjectMemberInvite.attributes);
       const projectId = _.parseInt(req.params.projectId);
       const invites = await models.ProjectMemberInvite.findAll({
         where: { projectId },
         raw: true,
       });
-      const opts = {
-        logger: req.log,
-        requestId: req.id,
-        memberFields,
-      };
-      const invitesWithDetails = await util.getObjectsWithMemberDetails(invites, fields, opts);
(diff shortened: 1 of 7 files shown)


src/util.js
```diff
@@ -443,6 +440,58 @@ _.assignIn(util, {
...
+  /**
+   * Filter member details by input fields
+   *
+   * @param {Array}   members   Array of member detail objects
+   * @param {Array}   fields    Array of fields to be used to filter member objects
+   * @param {Object}  opts      logger & request id
+   *
+   * @return {Array}            Filtered array of member detail objects
+   */
+  getObjectsWithMemberDetails: async (members, fields, opts) => {
```
maxceem: I would suggest simplifying the additional `opts` argument if possible. As currently, it has 3 variables inside, so to reuse this method we should research the source code of this method to understand what should be passed in this method.

I would suggest to remove the `opts` argument and pass the `req` only.
From the `req` we can get `logger` and `requestId `.

And regarding `memberFields` I guess we can implement this method without them, see comments bellow:

src/util.js
```diff
@@ -443,6 +440,58 @@ _.assignIn(util, {
...
+    if (!fields || _.isEmpty(fields)) {
+      return members;
+    }
+    const { logger, requestId, memberFields } = opts;
+    const requestedNonMemberFields = _.filter(fields, field => !_.includes(memberFields, field));
+    const otherMemberFields = ['photoURL', 'workingHoursStart', 'workingHoursEnd', 'timeZone'];
+
+    let allMemberDetails = [];
+    if (requestedNonMemberFields.length > 0) {
```
maxceem: I guess we can replace this condition with (for sure we can put the list to some constant):
```js
if (_.intersection(fields, ['photoURL', 'workingHoursStart', 'workingHoursEnd', 'timeZone','handle','firstName','lastName])) {
```

src/util.js
```diff
@@ -443,6 +440,58 @@ _.assignIn(util, {
...
+      const userIds = _.map(members, 'userId');
+      allMemberDetails = await util.getMemberDetailsByUserIds(userIds, logger, requestId);
+
+      if (_.intersection(requestedNonMemberFields, otherMemberFields).length > 0) {
```
maxceem: This conditions probably can be replaced with:
```js
if (_.intersection(fields, otherMemberFields).length > 0) {
```

src/routes/projectMemberInvites/list.js
```diff
@@ -0,0 +1,40 @@
...
+  permissions('projectMemberInvite.list'),
+  async (req, res, next) => {
+    try {
+      let fields = null;
+      if (req.query.fields) {
+        fields = req.query.fields.split(',');
+      }
+      const projectId = _.parseInt(req.params.projectId);
+      const invites = await models.ProjectMemberInvite.findAll({
+        where: { projectId },
```
maxceem: I've forgotten to mention that, in the invitation list we have to only return invites with status `pending`, same like we do for individual invite https://github.com/topcoder-platform/tc-project-service/blob/dev/src/models/projectMemberInvite.js#L68.

PR #415, opened 2019-11-28
Extended Local Setup

meshde:
# Local Docker Setup for tc-project-service

## Non-Topcoder Services

* The `docker-compose.yml` file under `local/full` extends the all services present in `local/docker-compose.yml` using the `extends` keyword.
* Apart from these, it also runs zookeeper and kafka services on ports `2181` and `9092` respectively.
* A separate PostgresSql instance is also run for `tc-notifictations`. No host port is mapped to this, and is only available internally by other services on port `5432`.
* A special kafka client service with the name `kafka-client` has been included. The sole purpose of this service is to create all topics listed under `local/full/kafka-client/topics.txt`. The service exists after all topics have been created.

## Topcoder Services

* All Topcoder services would be created extending from a generic base configuration, `local/full/generic-tc-service/Dockerfile`
* Containers for different services are spawned based on the build args `GIT_URL` and `GIT_BRANCH`
* The base configuration takes another build arg `BYPASS_TOKEN_VALIDATION`, which if set to `1` will add the required line to bypass token validation in `node_modules/tc-core-library-js/lib/auth/verifier.js`.
* Any service extending the base configuration can be forced to wait for a particular service/container to exit before starting it's main task. This has been used to make the kafka processing services wait for the`kafka-client` to create all required topics and exit. 
* All such services can connect with the host machine using the hostname `host.docker.internal`. This hostname works out of the box for the "Docker For Mac" application. To handle the case of Linux based hosts, an entry is made into `/etc/hosts` on conatiner startup such that it resolves to the host IP.
* There is also a base configuration for the `tc-notifications` service. Three services extend this base configuration, one for each of the three commands:
  * `npm run start`
  * `npm run startAPI`
  * `npm run reset:db`


maxceem:
FYI @vikasrohit now we have a way to run Connect App with Project Service locally pretty easily. This is available in v5 branch.

PR #423, opened 2019-12-13
[Hotfix] [PROD] es skills

maxceem:
### Overview
It happens that mapping of `details.taasDefinition.team.skills` in the ES index is defined as string, while in real form it's an array with objects. So when we try to add new projects to the index, ES throws an error:
```
[mapper_parsing_exception] failed to parse [details.taasDefinition.team.skills]
{\"type\":\"illegal_argument_exception\",\"reason\":\"unknown property [id]\"}
```

### Proper solution (long)
The only proper solution would need to make a long downtime of the service:
- find all existent projects with skills as string and update them to have an array instead
- remove index with the wrong mapping and recreate it from scratch, reindex all the projects now with fixed data, so mapping for skills would be set as array, and all new taas projects would be indexed successful

### Temporary solution (fast) - implemented in this PR
As a temporary workaround we would set  `details.taasDefinition.team.skills` to an empty string if it's defined before indexing it to ES. As in the current version on the details page we show data from DB, so if we don’t have skills in ES, the project details should be still displayed good.

### DIFF
--- a/src/events/projects/index.js
+++ b/src/events/projects/index.js
@@ -41,6 +41,10 @@ const indexProject = Promise.coroutine(function* (logger, msg) { // eslint-disab
       // removes non required fields from phase objects
       data.phases = data.phases.map(phase => _.omit(phase, ['deletedAt', 'deletedBy']));
     }
+    // TEMPORARY FIX: should fix ES mapping instead and reindex all the projects instead
+    if (typeof _.get(data, 'details.taasDefinition.team.skills') !== 'string') {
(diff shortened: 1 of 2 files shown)


RishiRajSahu:
@maxceem thanks for the quick changes, yep will need the proper solution by creating new index and reindexing. Also we can avoid downtime as the index name is configurable via env variables.

RishiRajSahu:
@maxceem working fine on dev after the feature branch deployed, pushing changes to prod.

PR #430, opened 2019-12-19
Syncing V5 changes into new develop branch

vikasrohit:
Now we should have develop branch pointing towards v5 work while dev is kept as reference for the last v4 work.


vikasrohit:
@maxceem fyi, merging v5 changes into develop branch as per our plan. Lets see if this works well and deploys to dev v5 cluster.

PR #438, opened 2020-01-10
[HOTFIX] [DEV] Create phase topics on project create

maxceem:


maxceem:
@vikasrohit previously we decided not to release this as a hotfix https://github.com/topcoder-platform/tc-project-service/pull/439 but include into a regular release. Though we didn't do so and this is not yet pushed to DEV.

I suggest pushing it to DEV and include it into the nearest release.

Btw, do you know any Project Template which causes creating phases on project creation? As I cannot find any for testing at the moment.

vikasrohit:
> Btw, do you know any Project Template which causes creating phases on project creation? As I cannot find any for testing at the moment.

https://connect.topcoder-dev.com/new-project/chatbot try this one, it should have the phases.

PR #439, opened 2020-01-10
[HOTFIX] [PROD] Create phase topics on project create

maxceem:


vikasrohit:
@maxceem I am closing the hotfix for this because for most the project templates on production, we don't create phases by default on project creation, so we are safe in terms of missing discussions for new projects. We would move this change to the production with our regular release.

PR #447, opened 2020-01-22
Fix for connect-app issue #3423

gets0ul:
https://github.com/appirio-tech/connect-app/issues/3423  

Enable admin user to join project as manager.

### DIFF
--- a/src/routes/projectMembers/create.spec.js
+++ b/src/routes/projectMembers/create.spec.js
@@ -9,7 +9,14 @@ import util from '../../util';
 import server from '../../app';
 import testUtil from '../../tests/util';
 import busApi from '../../services/busApi';
-import { USER_ROLE, BUS_API_EVENT, RESOURCES, CONNECT_NOTIFICATION_EVENT, INVITE_STATUS } from '../../constants';
+import {
+  USER_ROLE,
+  BUS_API_EVENT,
+  RESOURCES,
+  CONNECT_NOTIFICATION_EVENT,
+  INVITE_STATUS,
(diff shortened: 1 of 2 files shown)


src/routes/projectMembers/create.spec.js
```diff
@@ -201,6 +208,81 @@ describe('Project Members create', () => {
...
+        get: () => Promise.resolve({
+          status: 200,
+          data: {
+            id: 'requesterId',
+            version: 'v3',
+            result: {
+              success: true,
+              status: 200,
+              content: [{
+                roleName: USER_ROLE.MANAGER,
```
maxceem: As I understand this is a mocking request to get user roles https://github.com/topcoder-platform/tc-project-service/blob/develop/src/routes/projectMembers/create.js#L135.

The user we would call this is `testUtil.jwts.admin` who is admin, so it would get the role `USER_ROLE.TOPCODER_ADMIN` instead of `USER_ROLE.MANAGER`.

src/routes/projectMembers/create.spec.js
```diff
@@ -201,6 +208,81 @@ describe('Project Members create', () => {
...
+              }],
+            },
+          },
+        }),
+        post: () => Promise.resolve({
```
maxceem: Could you please point out which requests is mocked here?

src/routes/projectMembers/create.spec.js
gets0ul: Yes, that was a bad copy pasta. Will fix it.

gets0ul:
@maxceem PR is updated.

PR #457, opened 2020-02-03
change aws s3 cli to aws-sdk

yoution:


yoution:
@maxceem  please review again, I use `setTimeout` to do the delete action, please comment
the code  https://github.com/topcoder-platform/tc-project-service/blob/develop/src/routes/attachments/create.js#L73-L78 to validate

yoution:
thanks, your method is better then mine, `promise` is faster then `settimeout`

vikasrohit:
Thanks @yoution and @maxceem for the efforts.

vikasrohit:
@maxceem can we create hotfix instead of feature fix?

maxceem:
@vikasrohit I was thinking that we would make a release this week anyway, so created a feature branch. Also, it's better to test it on DEV a little.

Another thing, it looks like we have a redundant code there and unnecessarily call `upadlourl` File Service endpoint: https://github.com/topcoder-platform/tc-project-service/blob/feature/file-transfer-s3-improvements/src/routes/attachments/create.js#L73-L90 - I think we can just remove this call.

So after testing this fix, I was going to remove that call and also test without it.

Let me know if you still would like to have a hotfix for this. Should hotfix include teh second part with removing unnecessary call or no?

vikasrohit:
Makes sense. I was only worried if by any unavoidable reason, we have to push the release. Most probably we should be releasing on Thursday.  However, I think we do need some time in dev to test out the changes, so lets keep it feature for now and in case we don't have release this week on Thursday, we would make it hotfix.

maxceem:
@vikasrohit it looks like on DEV file uploading is not fully enabled. Files are visually uploaded but I cannot open them after. Could we enable file uploading on DEV for now, so we can fully validate it?

vikasrohit:
I think last time we made it enabled. I am not sure why it is not working now. I can check again, it was an env variable which used to prevent the actual file upload in dev.

PR #460, opened 2020-02-04
Feature/file transfer s3 improvements

maxceem:
2 improvements have been done to file transfer:
- use AWS SDK instead of AWS CLI vi exec for transferring files between buckets
- remove a redundant call for `${fileServiceUrl}uploadurl` which we don't need anymore in Project Service

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,7 +109,7 @@ workflows:
             - test
           filters:
             branches:
-              only: ['develop', 'feature/file-transfer-s3-improvements']
+              only: ['develop']
       - deployProd:
           context : org-global
           requires:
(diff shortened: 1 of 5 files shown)


maxceem:
@vikasrohit it has been tested on DEV and works good for me. So I propose to merge it to DEV for the release.

If we don't release, we can create a hotfix out of the `feature/file-transfer-s3-improvements` branch.

vikasrohit:
Just to double sure, we now don't call file service at all from project service, right?

maxceem:
We still call it to get a `downloadUrl` https://github.com/topcoder-platform/tc-project-service/blob/develop/src/routes/attachments/create.js#L100. We cannot avoid it.

Though now we don't have to call endpoint for getting `uploadUrl` anymore it left from the previous implementation.

vikasrohit:
Okay. I think we can avoid call for `downloadUrl` as well in near future, may be by having this as utility method in tc-core-lib-js.

PR #493, opened 2020-02-28
Feature/bulk milestone updates

panoptimum:
I fixed the wrong method argument as outlined in appeal to 1.1.1 at https://software.topcoder.com/review/actions/ViewReview?rid=493972
, the other point brought up by the reviewer of computed negative duration is invalid, because it's out of scope. I brought this up at https://apps.topcoder.com/forums/?module=Thread&threadID=951629&start=0 , but it was decided to keep that logic in the codebase.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ logs
 *.log
 npm-debug.log*
 .DS_Store
+.tern-port
+*#
 
 dist/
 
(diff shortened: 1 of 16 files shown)


maxceem:
Thank you for fixing the issue with the create milestone endpoint @panoptimum!

PR #503, opened 2020-03-11
Fix for issue #3740

tcchhabra:


tcchhabra:
checking

tcchhabra:
@maxceem I cannot use the other function, I have spent a good chunk of time try to reuse it but in vain so sticking with my function for handling this, which takes care of both the scenarios. Both test cases added.

PR #505, opened 2020-03-18
Use .env file for local setup

maxceem:
This PR is another step in making the local setup easier and reduce the mistakes which happen during local setup.

When we run Project Service locally we have to set environment variables for Auth0 two times:
- before running `docker-compose`
- before running `npm run start:dev`

It's very easy to forget to set them when restarting any of these commands.
- it's hard to notice when forget to set these variables so it takes time to debug before realizing this
- restarting `docker-compose` with env variables may be time-consuming
- I personally frequently come across this situation and so it regularly takes time from me

To reduce the chances of such mistake, this PR introduces using `.env` file so the workflow of working locally would be like this:
- create `.env` file in the project root directory (once)
- every time when we start `docker-compose` or `npm run start:dev` values from `.env` file would be always used. So even if we restart any of these commands in a new terminal window the values would be automatically used.
- `.env` file is ignored by Git


maxceem:
@vikasrohit could you please, have a quick look at this PR before merging it. In case you see any downsides and just FYI as you are also running Project Service locally.

vikasrohit:
@maxceem I am bit reluctant to have .env committed to the repo even though we have it excluded in .gitignore. I am kind of occupied with P0 things right now, so it might take few days to me to visit this again.

maxceem:
@vikasrohit I slightly changed README and removed `sample.env` file from the repo. I think it should reduce a chance for committing `.env` file with Auth0 values.

I propose to merge this PR and give it a try unit someone commits Auth0 values. If we found that someone does it due to using `.env` file, we would consider using another approach.

Btw, even without this approach, we had Auth0 values committed a few times before. Maybe if we explicitly provide a place where to define them, i. e. `.env` file which is ignored, it could even reduce the chance of mistake.

maxceem:
@vikasrohit could you, please, consider merging this PR again? I'd like to combine it with new demo data script https://github.com/topcoder-platform/tc-project-service/pull/542 and update README to simplify local setup guide.

vikasrohit:
@maxceem I don't see that .env is ignored in gitignore any more, I think we still need that so that accidentally people don't commit that file.

maxceem:
@vikasrohit it has been already ignored since long time ago https://github.com/topcoder-platform/tc-project-service/blob/develop/.gitignore#L5. As I've been using such a file locally since long time ago.

PR #516, opened 2020-03-19
Post-Processing Invites

gets0ul:
- change maskInviteEmails to postProcessInvites which will be post-processing on invite(s) with following constraints:
  - email field will be omitted from invite if the invite has defined userId
  - email field (if existed) will be masked UNLESS current user has admin permissions OR current user created this invite  
- also apply this post-processing when creating invite.

### DIFF
--- a/src/routes/projects/list.spec.js
+++ b/src/routes/projects/list.spec.js
@@ -64,6 +64,12 @@ const data = [
         email: 'test@topcoder.com',
         status: 'pending',
       },
+      {
+        id: 2,
+        email: 'hello@world.com',
+        status: 'pending',
+        createdBy: 1,
+      },
     ],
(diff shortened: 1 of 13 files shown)


src/routes/projects/list.spec.js
```diff
@@ -173,8 +173,8 @@ describe('LIST Project', () => {
   let project1;
   let project2;
   let project3;
-  before(function inner(done) {
-    this.timeout(10000);
+  before((done) => {
+    // this.timeout(10000);
```
maxceem: @gets0ul what is the reason to comment this line? I guess it was added to make sure unit test in other files are completed, and if we comment it, it could lead to unit test fail when we run all tests together.

PR #522, opened 2020-03-27
Fix random tests failures

xxcxy:


maxceem:
This fix has been done via F2F challenge http://www.topcoder.com/challenge-details/30120749/?type=develop.

It should make the unit test much more stable at least as per local testing FYI @vikasrohit.

vikasrohit:
Yep, I saw this PR and fixes looks good as I have done similar fix in past. The random errors are mostly because of sequences advancing and we still matching the records by hard coded id.

PR #530, opened 2020-04-08
Hotfix/user level reports

vikasrohit:


src/routes/projectReports/getEmbedReport.spec.js
```diff
@@ -210,10 +227,48 @@ describe('GET embed report', () => {
...
+          getUser.restore();
+          gem.restore();
+          cfg.restore();
+          if (err) {
+            done(err);
+          } else {
+            const resJson = res.body;
+            should.exist(resJson);
+            resJson.should.equal('generatedUrl');
+            const [user, project, member, embedUrl] = gem.lastCall.args;
```
maxceem, line 258: @vikasrohit I'm not sure how does it work. Here `gem.lastCall.args` is used but a few lines above on line `250` it's restored by `gem.restore()`.

vikasrohit: Yes, that is valid argument. However, I just copied the unit tests we already have for project reports route and updated them to meet my requirements. :)

vikasrohit: And strange thing is that, it is working fine. However, I did find one minor issue with unit tests of project reports route, they are still using magic numbers i.e. user ids are used as numbers instead of using them from the variables/constants and I fixed that in my unit tests.

maxceem:
I guess it's not critical, but we usually return responses in JSON, including situations when some error happens. In the report endpoints in case of errors responses are returned as plain text. Here are some examples and how they could be rewritten to return JSON.

```js
return res.status(500).send(err.toString());
```

can be replaced by:

```js
next(err);  // this would set HTTP status of error to `500` by default and format error JSON response properly
```

When some custom error code should be returned like for this case:
```js
return res.status(403).send('User is not allowed to access the report');
```

it can be replaced by:
```js
const err = new Error('User is not allowed to access the report');
err.status = 403;
return next(err);
```

src/routes/projectReports/getEmbedReport.spec.js
maxceem: Hey, @xxcxy could you please help us to clarify this moment how does it work?

vikasrohit:
> Other than returning strings instead of JSON in case of errors everything else looks good to me.

I thought of it but my thought process was because we return resources in response and for this endpoint the resource is actually the URL which is primitive data type instead of object. Still, we can think on it later and can fix it in a combined efforts.

maxceem:
> I thought of it but my thought process was because we return resources in response and for this endpoint the resource is actually the URL which is primitive data type instead of object. Still, we can think on it later and can fix it in a combined efforts.

Sure, it's not critical. Just for consistency, it's nice to know that we always get JSON in response, also to comply with Topcoder API V5 standard:

image

> We can fix the error response similarly later. Just curious, are we using that way of sending error responses in all other routes?

Yes, all other routes return JSON in case of errors (at least I don't any which don't).

Issue #492, opened 2020-02-28
Update node versions to the latest


maxceem:
We are currently using node v8 which is not longer maintained, see [Node maintanance schedule](https://nodejs.org/en/about/releases/)

![image](https://user-images.githubusercontent.com/146016/75510030-7c5ccb00-5a24-11ea-9644-3de2650dc9dc.png)

I would suggest updating to the latest version node v12.
It would be good to make this for all the repositories we are responsible for.

vikasrohit:
Is it something we can do in this release @maxceem ?

maxceem:
@vikasrohit I think yes. Though I gave it a try looks like the package `memwatch-next` that we use for detecting memory leaks doesn't support node 9+ and is abandoned https://github.com/marcominetti/node-memwatch/issues (latest version was released 4 years ago https://www.npmjs.com/package/memwatch-next).

Do we still use it and have to find a replacement first or we can remove it?

vikasrohit:
I don't think we need it now and at least immediately. But it might be helpful to find an alternative in parallel.

maxceem:
Done via PR https://github.com/topcoder-platform/tc-project-service/pull/531

codejamtc:
@maxceem @vikasrohit Please check. I have created a few projects using some user roles.
https://monosnap.com/file/ZBMlM4kjT9TqTMVOWUPEIqJz5Gzpkl

maxceem:
Thanks, @codejamtc. I think all good with this ticket.

codejamtc:
@maxceem 👍 I am unable to change the label, I think I don't have permission
You can change it to `Pass on Dev QA`

vikasrohit:
Closing. @codejamtc I am marking it as QA Pass for prod because the updated release is working fine and our standard smoke tests worked fine after the release.

PR #531, opened 2020-04-08
Upgrade to Node 12

maxceem:
Upgrade to Node 12 as per https://github.com/topcoder-platform/tc-project-service/issues/492
- removed `memwatch` as the package we use is not supported by Node 12, have to reimplement it again as per https://github.com/topcoder-platform/tc-project-service/issues/532
- removed redundant package `sleep` (not supported by Node 12) and all related unused code

Two general fixes have been done which could be included directly into `dev`:
- [fix: kafka consumer](https://github.com/topcoder-platform/tc-project-service/pull/531/commits/d3bc7cd1b5c47524b43d88e51b8d5e2bb54b7098)
- [fix: random test fail](https://github.com/topcoder-platform/tc-project-service/pull/531/commits/b45d4133bdb19ce71591c785007a8c997ab0ca03)

This PR is ready to be merged into `dev` so we can watch how does it behave in real-world usage (DEV env).


vikasrohit:
Nice catch about the random test failure for the phaseProduct. However, I didn't get the changes in Kafka consumer. Is that fix for unit tests only or is it related to one of the issues we are debugging around unordered processing of kafka messages which resulted in wrong invites data in ES?

vikasrohit:
@maxceem I am doing a squash merge for this PR instead of regular merge to make our life easier in case we have to revert the changes.

PR #547, opened 2020-04-14
Update database sequencers after importing

AliGebily:
Update database sequencers, after importing, to be set to max(autoincrement column) from table, so that when next insertions don't provide value of autoincrement column, as in case of using APIs, it should be set automatically based on last value of sequencers.


AliGebily:
Done, please check it.

Issue #483, opened 2020-02-21
Support standard M2M scopes


maxceem:
@vikasrohit I've got a request from the James Cori to support standard M2M scopes:
- `all:project`
- `read:project`
- `write:project`

Should we also keep the currently supported `all:connect_project` or can remove it?

Please, let me know if this task is good to proceed with or there are any decisions to make.

vikasrohit:
we will analyze and prioritize it.

rootelement:
Hi @vikasrohit , we're entering dev test phase for about the next 3 weeks. I'd appreciate if we could get this implemented soon in dev, so the production push of this could go in step with the challenge v5 api launch slated for the end of March. I don't want v5 going live with the all:connect_project scope if possible.

vikasrohit:
Sure @rootelement we will try to prioritize it.

@maxceem @rootelement can we add more details to what we expect (in terms of permissions) when these new scopes are associated with a m2m token? I mean `all:project` means all operations on all projects or all operations on the user's own projects. Similarly, `write:project` means updating only self (user is team member) projects or can update any project in system and so on. We need clear permission set for the scopes before we move ahead.

maxceem:
@vikasrohit as I understand M2M should get the level as of access as **admin** users, so:
- `read:project` means reading all the data in Project Service (as admins can do)
- `write:project` means being able to **create**, **update** and **delete** all the data in Project Service (as admins can do)
- `all:project` combination of two above

rootelement:
@maxceem is correct. This is a Machine to Machine token, so it will only be provisioned for our services to communicate. There's no "user" here. So what he has listed above is correct.

maxceem:
@vikasrohit should we keep support for the current scope `all:connect_project`? Or all the third party services which are using it now can update to a new `all:project` scope?

vikasrohit:
Okay, thanks for the updates @maxceem and @rootelement 

> should we keep support for the current scope all:connect_project? Or all the third party services which are using it now can update to a new all:project scope?

[...]

PR #555, opened 2020-04-26
Support M2M and Unified permissions, part 1

maxceem:
### Support of Standard M2M tokens (https://github.com/topcoder-platform/tc-project-service/issues/483)
3 endpoints sets are supported:
- `/projects`, scopes: `all:projects`, `write:projects`, `read:projects`
- `/projects/{id}/members`, scopes: `all:project-members`, `write:project-members`, `read:projcet-members`
- `/projects/{id}/invites`, scopes: `all:project-members`, `write:project-members`, `read:projcet-members`

Scope `all:connect_project` is supported for all endpoints for backward compatibility.

### Unified Permissions (as per https://github.com/topcoder-platform/tc-project-service/issues/346)
- For the same 3 endpoint sets use unified permissions rules:
   - all the permissions are defined in a declarative manner
   - all the permissions are defined in a single file https://github.com/topcoder-platform/tc-project-service/blob/feature/permissions/src/permissions/constants.js
- Added script to generate Permissions Documentation
   - so all the permissions can bee seen in a HTML file https://htmlpreview.github.io/?https://github.com/topcoder-platform/tc-project-service/blob/feature/permissions/docs/permissions.html
   - Added anchor link in the generate permissions document so we can link particular permissions or sections, for example [Project Invites Permissions](https://htmlpreview.github.io/?https://github.com/topcoder-platform/tc-project-service/blob/feature/permissions/docs/permissions.html#section-project-invite)

### Roles Matrix
- Refactored Create Project Member endpoint. Instead of imperative logic, the allowed project/topcoder roles associations are defined in a declarative way via constant [PROJECT_TO_TOPCODER_ROLES_MATRIX](https://github.com/topcoder-platform/tc-project-service/blob/feature/permissions/src/permissions/constants.js#L421)
- This matrix is added to be displayed in the permissions document https://htmlpreview.github.io/?https://github.com/topcoder-platform/tc-project-service/blob/feature/permissions/docs/permissions.html#roles-matrix

   image

### Fixes
- Fixed "initiatorUserId" and "userId" fields in 'connect.notification.project.team.updated' event payload
- Fix: don't allow to change project members Project Role if user doesn't have necessary Topcoder Roles. See how **Program Manager** which cannot join project as a Manager can still become Manager before this fix https://monosnap.com/file/SQ3nux0w9vKDsWcVYX6MSxW5Se5QKF
- Fix: don't allow copilots to manage Topcoder Team project members

### Features
- Allow directly create member by M2M and for admins

### TODO
- Don't return members in in `/projects` endpoints without `read:projcet-members` M2M scope

- It could be nice if the permissions which we now define in the Project Service regarding members and invites to apply in client side to make them absolutely in sync.
- Add more unit tests for M2M. At the moment I've added only one essential test.


src/permissions/index.js
```diff
@@ -3,35 +3,63 @@
...
-  Authorizer.setPolicy('project.viewMember', projectView);
-  Authorizer.setPolicy('project.removeMember', projectMemberDelete);
+  Authorizer.setPolicy('project.create', generalPermission(PERMISSION.CREATE_PROJECT));
+  Authorizer.setPolicy('project.view', generalPermission(PERMISSION.READ_PROJECT));
+  Authorizer.setPolicy('project.edit', generalPermission(PERMISSION.UPDATE_PROJECT));
+  Authorizer.setPolicy('project.delete', generalPermission(PERMISSION.DELETE_PROJECT));
+
+  Authorizer.setPolicy('projectMember.create', generalPermission([
+    PERMISSION.CREATE_PROJECT_MEMBER_OWN, // actually this permission includes the second permission and is enough
+    PERMISSION.CREATE_PROJECT_MEMBER_NOT_OWN,
```
vikasrohit, line 27: @maxceem `CREATE_PROJECT_MEMBER_NOT_OWN` should include `CREATE_PROJECT_MEMBER_OWN`, right? I mean a person who can create project member for others should be able to add project member for him/her-self.

src/util.js
```diff
@@ -869,23 +875,40 @@ _.assignIn(util, {
...
-        RESOURCES.PROJECT_MEMBER,
-        newMember);
 
-      return newMember;
+      // we have to remove all pending invites for the member if any, as we can add a member directly without invite
+      return models.ProjectMemberInvite.getPendingInviteByEmailOrUserId(member.projectId, null, newMember.userId)
+        .then((invite) => {
+          if (invite) {
+            return invite.update({
+              status: INVITE_STATUS.ACCEPTED,
```
vikasrohit, line 887: Assuming that this code is making invites accepted for the user who are directly invited to the project (e.g. by using m2m call to skip the invite process), I think we should not mark it as accepted. We should cancel it rather to indicate that user didn't accepted the invite, but he/she was directly adde the team.

maxceem: Agree. We should change it. NOTE: it was done like this before, I just moved the code to this method.

src/permissions/index.js
maxceem: @vikasrohit this works vise-verse. `CREATE_PROJECT_MEMBER_NOT_OWN` is stricter permission than `CREATE_PROJECT_MEMBER_OWN` so `CREATE_PROJECT_MEMBER_OWN` lets a lot of users to call this endpoint, while `CREATE_PROJECT_MEMBER_NOT_OWN` lets only some limited set of users to call it. So basically it would be enough to let calling this endpoint for `CREATE_PROJECT_MEMBER_OWN`. But I also list another permission `CREATE_PROJECT_MEMBER_NOT_OWN` just to keep it clear that such kind of users also can call this endpoint.

src/util.js
maxceem: @vikasrohit this is fixed now.

vikasrohit: > NOTE: it was done like this before, I just moved the code to this method.

Earlier we didn't have feature of skipping the invite process. So, I didn't understand when this piece of code was being called.

src/permissions/index.js
vikasrohit: okay, but I see it slightly different. I mean if I want to read the permission and deduce how that permission would work, then I guess, `CREATE_PROJECT_MEMBER_OWN` seems more restricted to me and `CREATE_PROJECT_MEMBER_NOT_OWN` less restricted. We might rename the later one to say `CREATE_PROJECT_MEMBER` to reflect the fact that it is a wider set of permission and allow a user who has this permission can do more operations than a person who have `CREATE_PROJECT_MEMBER_OWN` permission.

src/util.js
maxceem: Some Topcoder Members can join projects directly, like Managers. They could be also invited to the projects. Client side, if such user is invited, we block the project view and ask to accept or decline invite, so users cannot directly join if they are invited.

But API allow such users directly join project even if there is a pending invitation. So even though we don't have such situation in real life, API still allows it since the first implementation of invites, which I guess is fine.

vikasrohit: okay, got it. Basically, it was a rare case which not being hit so many times because it very rare to pending invite for a user, who is joining the project directly.

src/permissions/index.js
maxceem: That's an interesting view, I didn't watch it this way:
- my logic: wide permission means it allows wide range of users (but less operations)
- your logic: wide permission means it allow wide range of operations (but less users)

Your way sound like more logical for permissions.

It would be great to find a better naming, though me personally feeling kind of blind now for choosing a better naming after working on this non-stop. I have to step away a bit to refresh my mind from it.

I guess we would have to rename all the permissions (and [more](https://github.com/topcoder-platform/tc-project-service/blob/feature/permissions/src/permissions/constants.js)):

image

We have to come with some kind of convention in naming. If you see a clear naming convention for these permissions, we can rename them now.

Otherwise I would suggest to merge it now like this, in several days, when I refresh we can come up with a better naming approach.

vikasrohit: Lets merge for now.

Issue #560, opened 2020-05-04
Don't return members to the users who doesn't have permission to read them.


maxceem:
Follow up from https://github.com/topcoder-platform/tc-project-service/issues/483#issuecomment-609837470

At the moment we use M2M token to read members using `GET /project/{id}/members` endpoints. And we control WHO can read members using permission [READ_PROJECT_MEMBER](https://htmlpreview.github.io/?https://github.com/topcoder-platform/tc-project-service/blob/develop/docs/permissions.html#READ_PROJECT_MEMBER).

But we also return members in `GET /projects` and `GET /projects/{id}` endpoints.

To keep it consistent we should only return members with project object in these project endpoints if user has permission `READ_PROJECT_MEMBER`.

We should implement unit tests in each endpoint `GET /projects` and `GET /projects/{id}` to verify that members are not returned when we are making request using M2M token with `read:projects` scope, but without `read:project-members` scope.

PR #569, opened 2020-05-06
Fix for Issue #560 Don't return members to user without enough permission

gets0ul:
Only return members with project object in these GET /projects and GET /projects/{id} endpoints if user has permission READ_PROJECT_MEMBER.


gets0ul:
@maxceem PR is updated

Issue #561, opened 2020-05-04
For invites use scope "*:project-invites" instead of "*:project-members"


maxceem:
Follow up from https://github.com/topcoder-platform/tc-project-service/issues/483.

At the moment for invites we use the same M2M scopes as for members https://github.com/topcoder-platform/tc-project-service/issues/483#issuecomment-620409805. But as per https://github.com/topcoder-platform/tc-project-service/issues/483 we should have a separate scope for them:
- `all:project-invites`
- `write:project-invites`
- `read:project-invites`

maxceem:
@vikasrohit as we would have separate scope for invites, should we also stop returning them in `GET /projects` endpoints the same as we want to do for members https://github.com/topcoder-platform/tc-project-service/issues/560 for consistency?

vikasrohit:
Yes please.

maxceem:
Sum up:
- use separate scope for invites as per https://github.com/topcoder-platform/tc-project-service/issues/561#issue-611641836
- when we are returning invites in `GET /projects` or `GET /projects/{id}` endpoints we have to filter them:
  - **if** user has permission `READ_PROJECT_INVITE_NOT_OWN` then we can return in the list of all invites
  - **else if** user has permission `READ_PROJECT_INVITE_OWN` then we can return in the list only own invite (i. e. invite which has `userId` or `email` equal to the `req.authUser`)
  - **else**: user doesn't have any of the above permissions, so user gets empty list of invites

  create at least 1 unit test in each endpoint to check if M2M scope doesn't have `read:project-invites` then such a user wouldn't get any invites with project

  We already have this logic in the endpoint `GET /projects/{id}/invites`.

PR #570, opened 2020-05-06
Fix for Issue #561 Return empty invites for user without enough permission

gets0ul:
- use separate scope for invites
- return project with empty invites if user doesn't have enough permission

### DIFF
--- a/docs/permissions.html
+++ b/docs/permissions.html
@@ -636,8 +636,8 @@
 
               <div>
                   <span class="badge badge-dark" title="Allowed Topcoder Role">all:connect_project</span>
-                  <span class="badge badge-dark" title="Allowed Topcoder Role">all:project-members</span>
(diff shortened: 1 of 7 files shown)


gets0ul:
@maxceem PR is updated

gets0ul:
@maxceem fixed.

vikasrohit:
@maxceem I am pretty much sure that we have taken care of this but just want to be double sure, this change would not cause problem in returning invites for a user is invited to the project.

maxceem:
> @maxceem I am pretty much sure that we have taken care of this but just want to be double sure, this change would not cause problem in returning invites for a user is invited to the project.

Yes, we took care of it. And it's already deployed to DEV. So I guess during QA someone would notice if it accidentally got broken somehow. But as it's important functionality I would ask QA team to make regression testing here https://github.com/appirio-tech/connect-app/issues/4019.

PR #585, opened 2020-10-19
feat: remove RabbitMQ and use Kafka

maxceem:
- move actual handles from RabbitMQ to Kafka
- removed unused RabbitMQ handles code
- removed all RabbitMQ events firing
- added support for graceful shutdown of index-kafka.js
- removed RabbitMQ from production and local deployment
- fix unit tests to use a single ES service with index refresh fix
- removed redundant seedElasticsearchIndex.js script (we have a new one "npm run data:import")


vikasrohit:
Looks good to me. Merging to further test in dev.

PR #595, opened 2020-12-03
create job after project was created

imcaizheng:


src/events/projects/index.js
```diff
@@ -164,6 +187,40 @@ async function projectCreatedKafkaHandler(app, topic, payload) {
     await Promise.all(topicPromises);
     app.logger.debug('Topics for phases are successfully created.');
   }
+  if (project.type === 'talent-as-a-service') {
```
vikasrohit, line 190: @RishiRajSahu @maxceem please make sure this condition check would work. IMO, we should be looking at `templateId` instead of `project.type`. However, using `templateId` would require the condition check to be `contains` instead of `equals` and that list of templateId has to be updated with introduction of every new version of the template.
Other idea can be to use `alias` because we always try to keep the same alias for different versions of the same form. In that case we can have condition like `aliases.contains("talent-as-a-service")`.

vikasrohit (changes requested):
I think we need to fix the condition.

src/events/projects/index.js
maxceem: Thanks, @vikasrohit for keeping an eye on this. 

`project.type` is always set to the `projectTemplate.category` (which is actually a `ProjectType.key`) https://github.com/appirio-tech/connect-app/blob/feature/taas-jobs/src/projects/create/containers/CreateContainer.jsx#L251

See the list of project types here https://connect.topcoder-dev.com/metadata/projectTypes

So any project created by **ProjectTemplate** with **ProjectType**=**Talent as a Service** (`talent-as-a-service`) would be handled by this code. What do you think, can we use it to identify TaaS projects? 

IMO this is quite a good identifier unless we are going to use this **ProjectType**=**Talent as a Service** for some projects which shouldn't be handled by the **TaaS App**.

image

vikasrohit: Makes sense. I missed the `ProjectType` metadata. I guess, we are good for now.

vikasrohit (approved):

maxceem:
Merging for the further local testing and tweaks.

PR #596, opened 2020-12-04
[Kafka Consumer] Create TaaS Jobs during project creation

maxceem:
This PR contains:
- original PR https://github.com/topcoder-platform/tc-project-service/pull/595
- some small teaks to Job data
- token is temporary used from ENV variable because TaaS API doens't support M2M yet https://github.com/topcoder-platform/taas-apis/issues/40
- fix for the Kafka Topic Handlers config (would create a separate PR for this)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,7 @@ workflows:
             - test
           filters:
             branches:
-              only: ['develop']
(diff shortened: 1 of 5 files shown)


maxceem:
Deployed do DEV, fyi @RishiRajSahu

PR #597, opened 2020-12-04
[DEV] [HOTFIX] Kafka Topics Handlers Config

maxceem:
- As most of the topics have the same name, we cannot simply use them as objects keys. So we have to use a register method which would merge the config smartly.

The issue was caused by the fact, that multiple topics have the same name so when we defined several topics only the last one was really used. So only one topic wasn't processed:

image


maxceem:
Mering this PR for testing on DEV, fyi @vikasrohit .

Issue #620, opened 2021-02-08
New API endpoint to bring Billing Accounts available to the logged in user


vikasrohit:
For supporting the selection of appropriate billing account per project level, we need to provide list of available billing accounts to the logged in user so that he/she can choose from it and set it for the project. As per agreed decision after brainstorming with multiple stakeholders, we settled to implement this endpoint in the projects api itself instead of using v3 billing account api or creating separate api for calling SFDC api to bring the BA information.
This new endpoint would return all BA for the logged in user i.e. all the BAs that the logged in user is part of (as resource).

maxceem:
### Unit Tests to create:

- `should return 403 for anonymous user`
- `should return 403 for a regular user who is not a member of the project`
- `should return billing accounts for a regular user who is a member of the project`
- `should return all billing accounts to admin`
- `should return all billing accounts using M2M token with "read:user-billing-accounts" scope`

maxceem:
@vikasrohit could you please go through the test cases above and confirm if these are what is expected.

vikasrohit:
> should return all billing accounts for a regular user who is a member of the project

I think it should be available only to topcoder team. Customers won't be able to see available BAs.

maxceem:
ok, I've updated the comment https://github.com/topcoder-platform/tc-project-service/issues/620#issuecomment-776852879

PR #621, opened 2021-02-08
[DEV] New API endpoint to get Billing Accounts from SFDC

maxceem:
Draft PR to watch changes for https://github.com/topcoder-platform/tc-project-service/issues/620
fyi @vikasrohit


src/constants.js
```diff
@@ -274,6 +274,7 @@ export const M2M_SCOPES = {
     ALL: 'all:projects',
     READ: 'read:projects',
     WRITE: 'write:projects',
+    READ_BILLING_ACCOUNTS: 'read:user-billing-accounts',
```
maxceem, line 277: @vikasrohit I suggest renaming `READ_BILLING_ACCOUNTS` to something like `READ_USER_BILLING_ACCOUNTS` to highlight the difference from `WRITE_BILLING_ACCOUNTS: 'write:projects-billing-accounts'` as these scopes are for different things.

src/permissions/constants.js
```diff
@@ -91,6 +91,15 @@ const SCOPES_PROJECTS_WRITE = [
   M2M_SCOPES.PROJECTS.WRITE,
 ];
 
+/**
+ * M2M scopes to "read" available Billing Accounts for the project
+ */
+const SCOPES_PROJECTS_READ_AVL_BILLING_ACCOUNTS = [
+  M2M_SCOPES.CONNECT_PROJECT_ADMIN,
+  M2M_SCOPES.READ_BILLING_ACCOUNTS,
+  M2M_SCOPES.PROJECTS.ALL,
```
maxceem, line 100: @vikasrohit I guess adding `M2M_SCOPES.PROJECTS.ALL,` is not what is expected. `M2M_SCOPES.PROJECTS.ALL` allows to read/write general data of projects.

In this case we are adding restriction `M2M_SCOPES.READ_BILLING_ACCOUNTS` so this scope is required. If we also add `M2M_SCOPES.PROJECTS.ALL`, then M2M with `M2M_SCOPES.PROJECTS.ALL`, but without `M2M_SCOPES.READ_BILLING_ACCOUNTS` could read it.

So I guess we can remove `M2M_SCOPES.PROJECTS.ALL`.

src/permissions/constants.js
```diff
@@ -252,6 +261,19 @@ export const PERMISSION = { // eslint-disable-line import/prefer-default-export
...
+  /*
+   * Project Invite
+   */
+  READ_AVL_PROJECT_BILLING_ACCOUNTS: {
+    meta: {
+      title: 'Read Available Project Billing Accounts',
+      group: 'Project Billing Accounts',
+      description: 'Who can view the Billing Accounts available for the project',
+    },
+    topcoderRoles: ALL,
```
maxceem, line 273: @vikasrohit `topcoderRoles: ALL,` means that any user who is logged-in at Topcoder can view billing accounts.

Is it intentional, or would you prefer to use `projectRoles: ALL` which means, that only members of the project could get billing accounts of the project?

Then, we would also have to add `topcoderRoles: TOPCODER_ROLES_ADMINS,` so admins could read billing accounts for any project

src/routes/index.js
```diff
@@ -121,6 +121,9 @@ router.route('/v5/projects/:projectId(\\d+)/scopeChangeRequests/:requestId(\\d+)
   .patch(require('./scopeChangeRequests/update'));
 // .delete(require('./scopeChangeRequests/delete'));
 
+router.route('/v5/projects/:projectId(\\d+)/billingAccounts')
```
maxceem, line 124: @vikasrohit just would like to confirm. Do we get billing accounts available for user for the particular project? Or do we get accounts which are available for the user in ALL projects?

In the second case I guess we don't need to know the `projectId` and thus this endpoints might look like `/v5/projects/billingAccounts`.

maxceem (commented):
@vikasrohit I went through the code to understand the requirements here to write unit test and commented a few moments to confirm requriements.

src/constants.js
vikasrohit: Yep, that was exactly my intention as well to keep the new scope separate from reading billing account info on the project itself, missed or rather got confused with naming and left it as is. Go with `READ_USER_BILLING_ACCOUNTS`.

src/permissions/constants.js
vikasrohit: `M2M_SCOPES.PROJECTS.ALL` is for `connect:all_projects` (or some thing similar) scope? If yes, I think we should keep it to be backward compatible which it self is not super critical for this particular use case. If no, we can safely remove it.

src/constants.js
maxceem: Ok, good. Then for more clarification we would rename the existent `WRITE_BILLING_ACCOUNTS` to `WRITE_PROJECTS_BILLING_ACCOUNTS` to make them even more different.

src/permissions/constants.js
maxceem: `M2M_SCOPES.CONNECT_PROJECT_ADMIN = all:connect_project` which we use for backward compatibility is already here.

So yeah, `M2M_SCOPES.PROJECTS.ALL` can be removed.

src/permissions/constants.js
vikasrohit: Again it was because of my confusion with roles and my unit tests not working. Requirement is to allow any topcoder team member of the project to see the billing accounts available to them. I guess we should keep copilots as well, though I am not sure stakeholders want copilot to change the BA for a project. Also, admin (administrator and Connect Admin) should be able to view available billing accounts for any project.

src/routes/index.js
vikasrohit: It is the second case, however, the catch is that technically we can query available billing accounts for the given project as well. You can see some commented code in the actual route file. Another interesting fact is that ideally we wanted this endpoint to be in its own separate api (we didn't want to use the v3 billing accounts api because of its legacy origin) and we were not sure what that api should be, should it be v5 billing accounts api or a dedicated wrapper api over SFDC or a new endpoint in projects api itself (in case when projects api is owner of billing account data). So, we finally settled on the easiest option for now, implement a new endpoint in projects api.
And the reason I kept this endpoint under specific projects sub resources to have basic validation of a user being part of project to access this endpoint because I don't see any use of this end point as of now outside of the project's context.

maxceem: Ok, got it.

src/constants.js
vikasrohit: we can do that.

Issue #628, opened 2021-02-15
Create member endpoint should return member details


maxceem:
Endpoints `GET /v5/projects/:projectId/members`, `GET /v5/projects/:projectId/members/:id`, `PATCH /v5/projects/:projectId/members/:id` can return additional member details like 'photoURL', 'workingHourStart', 'workingHourEnd', 'timeZone', 'handle', 'email', 'firstName', 'lastName' if we explicitly list them in the `fields` query param. For example: `https://api.topcoder-dev.com/v5/projects/17076/members/?fields=id%2CuserId%2Crole%2CisPrimary%2CdeletedAt%2CcreatedAt%2CupdatedAt%2CdeletedBy%2CcreatedBy%2CupdatedBy%2Chandle%2CphotoURL%2CworkingHourStart%2CworkingHourEnd%2CtimeZone%2Cemail`

But the endpoint for creating new members `POST /v5/projects/:projectId/members` doesn't support returninig these additional fields. So if we want to get these additional details client-side, we would have to make 2 requests: create members and get member details.

We would like to update `POST /v5/projects/:projectId/members` to also support returning additional member details if they are explicitly listed in the `fields` query param.

Notes:
- we already have a reusable method to implement this `util.getObjectsWithMemberDetails` and we only have to properly apply it here
- it this method fails to get additional data, we still have to return response without additional data
- this method returns 'email', 'firstName', 'lastName' only to admins, which should works the same for this endpoint

Update Postman if needed.

maxceem:
This is done by https://github.com/topcoder-platform/tc-project-service/pull/629 and merged to DEV.

FYI @RishiRajSahu. We would need this for TaaS API https://github.com/topcoder-platform/taas-apis/issues/148

PR #629, opened 2021-02-15
Create member endpoint should return member details #628

xxcxy:


maxceem (approved):
Works good. Thank you @xxcxy

PR #657, opened 2021-07-25
[DEV] Milestone Management

elkhawajah:
This is part of Milestone management project/sprint for Topcoder conenct, an initiative to improve the UX in TC Connect for project progress/plan communication with the customers.

This push contains :
- Add new cross ref table between Topcoder Users and TC Connect Project Phase.
- Updated Phases api endpionts to include members field which represents the copilot user ids, so when pulling a phase or project phases we get members array
- Added new endpoint to update Phase copilots by passing userIds field to store the association
- Updated Get Phases endpiont to have memberOnly param to get only phases the authorized member is part of

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,7 +152,7 @@ workflows:
             - UnitTests
           filters:
             branches:
-              only: ['develop', 'connect-performance-testing', 'feature/shapeup_billing_accounts_protections']
+              only: ['develop', 'connect-performance-testing', 'feature/new-milestone-concept']
       - deployProd:
           context : org-global
           requires:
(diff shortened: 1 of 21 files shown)


RishiRajSahu:
@elkhawajah before merging we have to  get this updated as per the review from @maxceem 
-----
it looks like this new code would return phase members together with PII information like first name, last name and email to any user
while previously we restricted such info only to admins
there is already a reusable method util.getObjectsWithMemberDetails which can populate member details based on who made the req here are the examples:
https://github.com/topcoder-platform/tc-project-service/blob/feature/new-milestone-concept/src/routes/projectMembers/get.js#L89
https://github.com/topcoder-platform/tc-project-service/blob/feature/new-milestone-concept/src/routes/projectMembers/list.js#L105

PR #753, opened 2023-09-07
use the new file scanner service to av scan attachments

ThomasKranitsas:


Aranyajit:
The changes are good.

The Kafka handlers registration can be simplified to follow an uniform way to avoid any consfusion

PR #774, opened 2025-01-26
feat(PM-571): create copilot request endpoint

hentrymartin:
## What's in this PR?

- Created an endpoint to create copilot request
- Added the endpoint in the swagger yaml file

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -24,7 +24,7 @@ securityDefinitions:
     type: apiKey
     name: Authorization
     in: header
-paths:
+paths:              
   "/projects":
     get:
       tags:
@@ -244,6 +244,38 @@ paths:
           description: Internal Server Error
           schema:
             $ref: "#/definitions/ErrorModel"
+  "/projects/{projectId}/copilot-request":
+    post:
+      tags:
+        - projects copilot request
+      operationId: createCopilotRequest
+      security:
+        - Bearer: []
+      description: "create copilot request"
+      responses:
+        "200":
+          description: Created copilot request
+          schema:
+            $ref: "#/definitions/Copilot"
+        "401":
+          description: Unauthorized
+          schema:
+            $ref: "#/definitions/ErrorModel"
+        "400":
+          description: Bad request
+          schema:
+            $ref: "#/definitions/ErrorModel"
+        "500":
+          description: Internal Server Error
+          schema:
+            $ref: "#/definitions/ErrorModel"
+      parameters:
+      - $ref: "#/parameters/projectIdParam"
+      - name: body
+        in: body
+        required: true
+        schema:
+          $ref: "#/definitions/NewCopilotRequest"
   "/projects/{projectId}/attachments":
     get:
       tags:
@@ -5459,6 +5491,14 @@ definitions:
         type: string
       address:
         type: string
+  NewCopilotRequest:
+    type: object
+    required:
+      - data
+    properties:
+      data:
+        type: object
+        description: copilot request data
   NewProject:
     type: object
     required:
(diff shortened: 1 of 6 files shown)


src/models/copilotRequest.js
kkartunov, line 1: Please make model more reliable:
- status should not allow null, limit the string to max 16 length,  put/use a validate function to check against `COPILOT_REQUEST_STATUS`
- data should not allow null
- where is `projectId` reference?

src/routes/copilotRequest/create.js
```diff
@@ -0,0 +1,72 @@
...
+
+module.exports = [
+  validate(addCopilotRequestValidations),
+  (req, res, next) => {
+    const data = req.body;
+    if(!util.hasPermissionByReq(PERMISSION.MANAGE_COPILOT_REQUEST, req)) {
+      const err = new Error('Unable to create copilot request');
+      _.assign(err, {
+        details: JSON.stringify({ message: 'You do not have permission to create copilot request' }),
+        status: 400,
```
kkartunov, line 24: Should be 403.

src/routes/copilotRequest/create.js
```diff
@@ -0,0 +1,72 @@
...
+    const projectId = _.parseInt(req.params.projectId);
+    _.assign(data, {
+      projectId,
+      status: COPILOT_REQUEST_STATUS.NEW,
+      createdBy: req.authUser.userId,
+      updatedBy: req.authUser.userId,
+    });
+
+    models.sequelize.transaction((transaction) => {
+      req.log.debug('Create Copilot request transaction');
```
kkartunov, line 38: I think it makes sense to dump also the request data here for logging purposes.

src/routes/copilotRequest/create.js
```diff
@@ -0,0 +1,72 @@
...
+        where: { id: projectId, deletedAt: { $eq: null } },
+      })
+        .then((existingProject) => {
+          if (!existingProject) {
+            const err = new Error(`active project not found for project id ${projectId}`);
+            err.status = 404;
+            throw err;
+          }
+          return models.CopilotRequest.findOne({
+            where: {
```
kkartunov, line 49: Please extend the logic, NEW, APPROVED, SEEKING should block creating new request. The others should allow creating new copilot request.

src/routes/index.js
```diff
@@ -378,6 +378,10 @@ router.route('/v5/projects/:projectId(\\d+)/settings')
   .get(require('./projectSettings/list'))
   .post(require('./projectSettings/create'));
 
+// Project Copilot Request
+router.route('/v5/projects/:projectId(\\d+)/copilot-request')
```
kkartunov, line 382: I am fine having the `projectId` in the URL. Need to comunicate this to @himaniraghav3.
Please change `/copilot-request` to `/copilots/request`.

kkartunov (changes requested):
- described above some items to modify
- there is no migration included, please create such for creating the new table `copilot_requests` and link it to execute as the rest.

src/models/copilotRequest.js
hentrymartin: @kkartunov projectId is referenced/associated inside Project model. I've updated the PR based on other comments!

hentrymartin:
@kkartunov Do we need a migrations? because migrations/sync.js automatically creates the table based on the defined model.

himaniraghav3 (commented):
Looks good as long as Kiril's comments are addressed
Also, I think there needs to be an error response for an existing request - "The copilot request already exists"

kkartunov:
> @kkartunov Do we need a migrations? because migrations/sync.js automatically creates the table based on the defined model.

yes, that script is running when resetting db https://github.com/topcoder-platform/tc-project-service/blob/13140b97aad9e6b8319080080c2910b7848320e0/package.json#L19. Should not happen by default.

docs/swagger.yaml
```diff
@@ -244,6 +244,38 @@ paths:
...
+        - projects copilot request
+      operationId: createCopilotRequest
+      security:
+        - Bearer: []
+      description: "create copilot request"
+      responses:
+        "200":
+          description: Created copilot request
+          schema:
+            $ref: "#/definitions/Copilot"
```
himaniraghav3, line 259: I think you missed adding the Copilot definition

himaniraghav3: Oh it's just misspelled

jmgasper (commented):
In general, I really dislike shoving JSON into a column in the database.  It makes it a lot  harder to run DB queries to pull data from the column - is there any reason we just have a blanket `data` column instead of splitting it into individual columns for individual fields?

PR #778, opened 2025-02-05
PM-571 Validate data in create copilot request endpoint

himaniraghav3:
### What's in this PR?

We decided to dump all the request copilot form data inside data because we're not sure of what the fields will look like. 
For now, we are adding validation to the data in the endpoint 
Also added it to swagger


jmgasper (commented):
The project ID should be a number, correct?  Currently it's numbers in the DB / API.

kkartunov (approved):

PR #780, opened 2025-02-17
PM-588 copilot requests

vas3a:
- Add new endpoint: - get copilot request by id
- updated the path for get copilot requests to allow getting all requests without specifying projectId


kkartunov (approved):
Looks good.

PR #789, opened 2025-03-04
PM-589 Make opportunity routes public

himaniraghav3:
### What's in this PR?

Make the Opportunity listing and details APIs public.


himaniraghav3:
I changed the get by id endpoint to be singular while we're at it

kkartunov (approved):

PR #808, opened 2025-05-19
PM- 1207 Copilot Opportunity details api is not public

himaniraghav3:
https://topcoder.atlassian.net/browse/PM-1207

The v5/projects/copilot/opportunity/:id api wasn't being read as a public route. 
This fix ensures that it's public


kkartunov (approved):
Looks good.

kkartunov (changes requested):
@himaniraghav3 why we do have/use different paths?
```
/projects/copilots
/projects/copilot
```
can we get it all consistent, either use one or the other?

PR #809, opened 2025-05-19
fix(PM-1168): invite user when assigning as copilot

hentrymartin:
## What's in this PR?

-  invite user when assigning as copilot

Ticket link - https://topcoder.atlassian.net/browse/PM-1168

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,7 +149,7 @@ workflows:
           context : org-global
           filters:
             branches:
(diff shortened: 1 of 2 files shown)


kkartunov (changes requested):
We should handle only of the invitation is accepted by the copilot. As currently implemented this directly accepts and modifies records which should be done only after invitation is accepted.

kkartunov (approved):
Looks good.

PR #810, opened 2025-05-23
fix: QA feedbacks for assigning copilots

hentrymartin:
## What's in this PR?

- Fixed QA feedbacks when assigning a copilot to an opportunity.


kkartunov (approved):
Looks good.

PR #814, opened 2025-05-28
fix(PM-1169): Send source as part of email resource

hentrymartin:
## What's in this PR?

- Send source as part of email resource


kkartunov (changes requested):

kkartunov (approved):
Looks good

